PR #1, opened 2015-10-08
move from env vars to labels

lloesche:
Use labels instead of env vars.


discordianfish:
I've cherry-picked the commit. Thanks!

PR #3, opened 2015-11-02
Load templates only once

felixb:
IMHO it is overkill to load templates from disk everytime the config is generated.


discordianfish:
Makes sense, thanks!

PR #8, opened 2015-11-10
Marathon-LB enhancements.

brndnmtthws:
- Respect health checks by default
 - Supply some better default templates
 - Implement workaround for Alpine Linux DNS issue (it does not respect
   ordering of DNS servers, and we include a non-Mesos-DNS entry on
   DCOS)

cc @discordianfish


templates/HAPROXY_BACKEND_HTTP_HEALTHCHECK_OPTIONS
```diff
@@ -0,0 +1,2 @@
+  option  httpchk GET {healthCheckPath}
```
guenter: Why have the load balancer do health checks in addition to Marathon?

brndnmtthws: Good question. I guess the concern would be what happens when there is a disconnect between Marathon, the Python script, and HAProxy which results in leaving dead backends in the pool. For example, in the case of a network partition.

run
```diff
@@ -10,6 +10,10 @@ else
   rm /tmp/server-*.pem
 fi
 
+# This is a hack due to musl libc's DNS handling
+cat /etc/resolv.conf | head -n 3 > dns
+cat dns > /etc/resolv.conf
```
discordianfish: This would always take the first three lines, nobody guarantees that it includes any dns server..
But good point, I just ran into a related issue and also realize that this is a problem on DCOS. We probably have to change this to something glibc based.. Either way, let's address this in a separate PR.

brndnmtthws: Yeah, that's intended. If it's a cluster with a single master, than only the first DNS server is valid. I'll move it to another PR.

discordianfish:
Looks good in general, although we need to find a different workaround for the musl issue.
Either we convert this to debian or we create our own base image. In that case I would fork https://github.com/sdurrheimer/alpine-glibc which is what the prometheus team is using as well and so far we didn't have problems wit that.

Regarding the templates: If we feel like those are more reasonable defaults, I would suggest to change the defaults in the script.

run
discordianfish: Right, but there could be no comment or more than 2 lines comment in the file, so we can do it like this.

brndnmtthws:
I split out the DNS change.

templates/HAPROXY_HEAD
```diff
@@ -0,0 +1,18 @@
...
+  tune.ssl.default-dh-param 2048
+defaults
+  log               global
+  retries           3
+  maxconn           2000
+  timeout connect   5s
+  timeout client    50s
+  timeout server    50s
+listen stats
+  bind 0.0.0.0:9090
```
discordianfish: Deployed to a public slave, this will make the stats accessible to the public internet.
But it's a good point, not sure how to properly solve that. What I actually want here is to map this port to the hosts *internal* interface but I don't think this is possible with marathon.

brndnmtthws: I think in "the real world" a public slave wouldn't necessarily be accessible from the internet. You'd probably have an ELB in front of it. We actually already do this with AWS (we have a public slave ELB, although it's not properly configured).

brndnmtthws:
On another note: what should we do about reserving ports? We should probably include port 9090 in the marathon.json, and perhaps also a range of ports to which we might attach services.

templates/HAPROXY_HEAD
discordianfish: Well, in our OVH cluster we don't have a LB in front of it. I would leave it on 127.0.0.1. If you need to access it, you can schedule whatever needs to read from that to the public slave as well.

discordianfish:
@brndnmtthws Reserving ports is definitely an issue. Right now marathon-lb binds to the 'service ports' and it can't know them when scheduled, so it's very prone to port conflicts. We can disable that and have it only listen on 80/443 for incoming requests but then people can't use the service ports. So that's a product decision.

templates/HAPROXY_HEAD
brndnmtthws: I think it would be great to just make it really easy to supply alternate templates. One idea might be to include templates as URIs, and automatically copy them from the sandbox into the `templates/` dir at startup. Thoughts?

brndnmtthws:
Could we be a little more prescriptive about port management? At the very least, we could make a recommendation (i.e., start numbering your service ports from port 10000 or something).

discordianfish:
@brndnmtthws Yes, we could do that. Allocate 10000-20000 in marathon-lb and advise people to only use service ports in that range.

brndnmtthws:
Unfortunately Marathon doesn't let you reserve a range of ports, so we'd have to iterate 10k ports in the JSON. Maybe we'll go with 10000-10100, and if someone needs more ports they can add them :)

discordianfish:
Oh.. Hoped that would be possible. Well in that case I would probably just ignore this for now - or disable the service ports by default. To properly solve this we need a framework dynamically allocating the service ports on the public slaves.

templates/HAPROXY_HEAD
discordianfish: Not sure how this would look like. You mean something that would actually fetch that from http urls? Or is there a way to add files to the sandbox from a universe package somehow?

brndnmtthws: In the `run` script we'd do something like:

`cp $MESOS_SANDBOX/templates/* templates/`

And then we'd make it easier to configure the Marathon config to specify additional URIs.

Does that make sense?

discordianfish: That's reasonable but I don't understand how the templates get into the sandbox in the first place. Can you point me to the feature we would use for that?

brndnmtthws: If you specify URIs in the Marathon JSON, Mesos will download the artifacts and place them into the Mesos sandbox. Mesos will then export the $MESOS_SANDBOX env var inside the Docker container, and mount the sandbox as a volume at that path.

See: https://mesos.apache.org/documentation/latest/docker-containerizer/ under the section "What does the Docker Containerizer do?"

templates/HAPROXY_BACKEND_HTTP_HEALTHCHECK_OPTIONS
discordianfish: So if we think this makes sense we should just make those defaults. Thoughts?
@brndnmtthws Maybe you can split out the `--health-check` changes and create a separate PR for that? IMO would make sense to have this in a separate commit anyway

guenter: synced with @brndnmtthws on this requirement - the check should be done from the LB because there could be a network partition where marathon can access the service but the LB can't

discordianfish: :+1: Sounds reasonable.
@brndnmtthws: Can you move the template changes to the defaults in the script? I think if this is how it's support to be in DCOS by default, it should also be default in the script.

brndnmtthws:
Updated the PR. Removed the templates, and instead set those as defaults.

scalp42:
@brndnmtthws @discordianfish any chance to see this merged this week?

discordianfish:
@brndnmtthws ^- Can you look into this?

brndnmtthws:
@scalp42 Interesting. I don't see how the changes in this PR might affect that. I did notice, however, that the logs from the 2 examples above are slightly different. Namely, the first one attempts to reload HAProxy, whereas the second does not (`servicerouter: reloading using /etc/init.d/haproxy reload`...)

scalp42:
Right @brndnmtthws I didn't paste the full log, the reloading is working fine in both cases.

It's just the iteration of apps that doesn't work anymore between master and your branch (which I would need for the health checks from Haproxy).

I'm hoping to take a look at the code today.

brndnmtthws:
Okay, fair enough. I'm going to test it again to make sure I didn't break something.

scalp42:
Sounds good @brndnmtthws I'll report what I find as well (hoping before noon).

Thanks for the work also!!

scalp42:
Okay I found the issue, if the app doesn't have health checks then it gets skipped for whatever reason.

Here is requestbin1:

```json
{
  "container": {
    "type": "DOCKER",
    "docker": {
      "network": "BRIDGE",
      "image": "picoded/httpbin",
      "portMappings": [{"containerPort": 80, "hostPort": 0, "protocol": "tcp"}]
    }
  },
  "id": "requestbin1",
  "instances": 1,
  "cpus": 0.1,
  "mem": 32,
  "labels": {
    "HAPROXY_0_MODE": "http",
    "HAPROXY_0_VHOST": "requestbin1"
  },
  "healthChecks": [
    {
        "protocol": "HTTP",
        "path": "/delay/2",
        "gracePeriodSeconds": 3,
        "intervalSeconds": 10,
        "portIndex": 0,
        "timeoutSeconds": 5,
        "maxConsecutiveFailures": 3
    }
  ]
}
```

Here is `requestbin2`:

```json
{
  "container": {
    "type": "DOCKER",
    "docker": {
      "network": "BRIDGE",
      "image": "picoded/httpbin",
      "portMappings": [{"containerPort": 80, "hostPort": 0, "protocol": "tcp"}]
    }
  },
  "id": "requestbin2",
  "instances": 1,
  "cpus": 0.1,
  "mem": 32,
  "labels": {
    "HAPROXY_0_MODE": "http",
    "HAPROXY_0_VHOST": "requestbin2"
  }
}
```

I modified the part here https://github.com/brndnmtthws/marathon-lb/blob/brndnmtthws/lb-improvements/marathon-update-haproxy.py#L1005-L1010 with the following code:

```python
    # Convert into a list for easier consumption
    apps_list = list()
    for marathon_app in marathon_apps:
        logger.info("MARATHON_APP=%s" % marathon_app.services)
        for service in marathon_app.services.values():
            logger.info("SERVICE=%s" % service)
            if service.backends:
                logger.info("SERVICE_BACKENDS=%s" % service.backends)
                apps_list.append(service)
    logger.info("APPS_LIST=%s" % apps_list)
    return apps_list
```

When running with your branch and the modified code with extra logging:

```
servicerouter: MARATHON_APP={10000: MarathonService(u'/requestbin1', 10000)}
servicerouter: SERVICE=MarathonService(u'/requestbin1', 10000)
servicerouter: SERVICE_BACKENDS=set([MarathonBackend(u'192.168.99.112', 31393)])
servicerouter: MARATHON_APP={10001: MarathonService(u'/requestbin2', 10001)}
servicerouter: SERVICE=MarathonService(u'/requestbin2', 10001)
servicerouter: APPS_LIST=[MarathonService(u'/requestbin1', 10000)]
```

You can see that `requestbin2` does not get appended to the `apps_list` list as the `if service.backends` skips it.

You can see the output from Marathon apps endpoint for both requestbin1 and requestbin2 apps: https://gist.github.com/scalp42/5163046000e9ed1bc055

brndnmtthws:
Ah yes. I know why this is. I'll update the PR in a sec.

scalp42:
Haha @brndnmtthws amazing, I updated my comment with more details in case.

scalp42:
Also if I could make a suggestion please add the following to `.gitignore`:

```
templates/HA*
```

But still create the empty dir `templates` with `.gitkeep` inside it if it makes sense.

This is for the folks **not** using the Docker container (and makes it configuration management friendly).

scalp42:
My bad, `templates/HA*` is less preferred to `templates` but it's nitpicking at this point.

Thanks a lot for your work! Going to test it now.

scalp42:
Working perfectly :100:

scalp42:
@brndnmtthws still happening :/

Let me try to figure it out as well.

brndnmtthws:
Hmm, that might not be the intended behaviour. The idea behind `--health-check` is to use Marathon's health checks. The code you pointed to above is if you want to enable HAProxy's health checks independent of Marathon's.

I'm not suggesting that's intuitive, but that's currently how it's implemented.

brndnmtthws:
`--health-check` only applies to _Marathon's_ health checks. In other words, if it's enabled, the app endpoints (backends) will only be exposed when Marathon reports the app instances as being healthy. If there are no health checks set in Marathon, the app's backends will be added to the LB pool regardless of their health state.

You can _still_ use HAProxy's health checks regardless of Marathon's status, which is where the templates come into play.

Is that more clear?

I'll also update the `--help` info to clarify.

scalp42:
Oh that makes way more sense now gotcha:

- the health checks specified through Marathon are **always** translated to Haproxy health checks **unless** you override `HAPROXY_BACKEND_HTTP_HEALTHCHECK_OPTIONS` to blank for example

- when `--health-check` is specified, if Marathon fails a health check to one of the app, then it gets removed from the configuration file itself in Haproxy.

:+1:

brndnmtthws:
That's right!

:bomb:'s away!

scalp42:
Huge :+1: again, much much appreciated.

PR #9, opened 2015-11-10
DNS workaround for DCOS w/ Alpine Linux.

brndnmtthws:
Alpine Linux uses musl libc which does not respect the order of DNS
servers in /etc/resolv.conf. Since we include a fallback DNS server as
the last entry (which is not Mesos-DNS), DNS lookups will randomly fail.

cc @discordianfish

### DIFF
--- a/run
+++ b/run
@@ -10,6 +10,10 @@ else
   rm /tmp/server-*.pem
 fi
 
+# This is a hack due to musl libc's DNS handling
+cat /etc/resolv.conf | head -n 3 > dns
+cat dns > /etc/resolv.conf
+
 /sbin/runsv "$SERVICE" &
 RUNSV=$!
 


discordianfish:
Okay, let's discuss the workaround here: The workaround doesn't work reliable since the resolv.conf could look different and we can't just expected that we need the first 3 lines.
Instead we could:

1) Use debian base image
2) Use alpine with glibc, like https://github.com/sdurrheimer/alpine-glibc
3) Parse resolv.conf properly, including comments etc

I tend towards 2), basically we would just create mesosphere/docker-base or something based on sdurrheimer's image.

brndnmtthws:
Option 2 sounds great to me! I'll test it out with this one: https://hub.docker.com/r/sdurrheimer/alpine-glibc/

brndnmtthws:
It looks like simply adding glibc doesn't help. The python binaries still link to musl libc and I can't seem to make it load glibc instead (maybe PEBKAC?).

discordianfish:
You're right, this only works if you compile things after installing glibc.. I've did a new PR to change all this to debian.. Guess it's still the best option: https://github.com/mesosphere/marathon-lb/pull/10

discordianfish:
Ok this can be closed now that we switched to debian.

PR #10, opened 2015-11-11
Use debian as base image and bash for scripts

discordianfish:
Alpine is much smaller but it suffers from severe DNS issues. It lacks
support for search domains and tries to resolve a dns name via all
configure nameservers and returns the first answer. Since we keep a
non-mesos dns server for backup, sometimes this one is faster and we get
NXDOMAIN for internal names.

- https://github.com/gliderlabs/docker-alpine/issues/8
- https://github.com/gliderlabs/docker-alpine/issues/11

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,12 @@
-FROM alpine
+FROM debian
 
 ENTRYPOINT [ "/marathon-lb/run" ]
 CMD        [ "sse", "-m", "http://leader.mesos:8080" ]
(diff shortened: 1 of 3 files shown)


brndnmtthws:
Looks great to me. Has it been tested?

discordianfish:
Yes I did some manual testing. Will deploy it to our OVH cluster to test it some more after merging.

SydOps:
This is embarrasses that alpine Linux need helps from Debian.

PR #11, opened 2015-11-12
Support reading templates from mesos sandbox

discordianfish:
@brndnmtthws: Like this?


brndnmtthws:
Yes, exactly :)

:+1:

PR #12, opened 2015-11-16
Add --skip-reload to not reload Haproxy on changes

scalp42:
cc @discordianfish

### DIFF
--- a/marathon-update-haproxy.py
+++ b/marathon-update-haproxy.py
@@ -927,10 +927,13 @@ def compareWriteAndReloadConfig(config, config_file):
 
     if runningConfig != config:
         logger.info(
-            "running config is different from generated config - reloading")
+            "running config is different from generated config...")
         writeConfig(config, config_file)
-        reloadConfig()
-
+        if not args.skip_reload:
+            logger.info("...reloading Haproxy.")
+            reloadConfig()
+        else:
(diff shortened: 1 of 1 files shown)


discordianfish:
@scalp42 What's the use case here? Can you explain why you need this option?

scalp42:
One of the use case is deploying an app through Marathon but not wanting to have the loadbalancing layer to immediately pick up the change (say you want to test an app on the slave locally but don't want Haproxy to pick it up).

brndnmtthws:
Not sure I see the point, to be honest. I'm not completely opposed to it, however.

Also, can you separate the style changes out into a separate PR?

scalp42:
@brndnmtthws should be good now :smile:

brndnmtthws:
LGTM. Any comments @discordianfish?

brndnmtthws:
Is this still something you need? I'm going to close the PR for now, but feel free to reopen if you want to discuss.

scalp42:
@brndnmtthws unfortunately I still need it as it allows to "dump/debug" by launching the service twice and passing another location for `haproxy.cfg` but *making sure* it does not actually try to start/reload Haproxy.

scalp42:
I can rebase my changes in case!

PR #13, opened 2015-11-17
Add --disable-service-port flag

discordianfish:
This flag will prevent the script from adding frontends for service
ports, unless there is a explicit HAPROXY_{n}_ENABLE_SERVICE_PORT label
set.

--
This isn't done, ran into some haproxy error. Will fix and let you know when fixed. Early feedback welcome though.

### DIFF
--- a/marathon-update-haproxy.py
+++ b/marathon-update-haproxy.py
@@ -484,6 +484,8 @@ def set_port(x, y):
 def set_mode(x, y):
     x.mode = y
 
+def enable_service_port(x, y):
+    x.enable_service_port = True
 
 label_keys = {
     'HAPROXY_{0}_VHOST': set_hostname,
@@ -492,7 +494,8 @@ label_keys = {
     'HAPROXY_{0}_SSL_CERT': set_sslCert,
     'HAPROXY_{0}_BIND_ADDR': set_bindAddr,
     'HAPROXY_{0}_PORT': set_port,
-    'HAPROXY_{0}_MODE': set_mode
+    'HAPROXY_{0}_MODE': set_mode,
+    'HAPROXY_{0}_ENABLE_SERVICE_PORT': enable_service_port
 }
 
 logger = logging.getLogger('servicerouter')
@@ -513,7 +516,8 @@ class MarathonBackend(object):
 
 class MarathonService(object):
 
-    def __init__(self, appId, servicePort, healthCheck):
+    def __init__(self, appId, servicePort, healthCheck, args):
+        print(args)
         self.appId = appId
         self.servicePort = servicePort
         self.backends = set()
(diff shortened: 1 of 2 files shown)


scalp42:
Not related but I'm confused about the HAPROXY_{0}_XXX labels.

If I try to set one the app with the labels HAPROXY_1_xx, Haproxy is not picking up for example the TCP mode if set.

Only HAPROXY_0_XXX labels get picked up.

Any idea?

discordianfish:
@scalp42 That's the point of this number/id. HAPROXY_1_xx will be only picked up by marathon-lb instances running with `--group 1`. If that's not working as expected, please open a new issue.

discordianfish:
@brndnmtthws Okay, this should work now. The docker image now disables the service port exposition by default via the introduced flag. I've tested it locally and it looks good.

brndnmtthws:
Not quite what I had in mind for this implementation, but it's still useful.

In this case, maybe it should be on by default.

I want a separate label (like `LB_TAGS="internal,external"`) for controlling which LBs expose the services. Each LB will have a corresponding tag associated with it.

scalp42:
@discordianfish what does this `--group` param refers to ?

marathon-update-haproxy.py
```diff
@@ -1102,14 +1109,23 @@ def get_arg_parser():
...
+    parser.add_argument("--disable-service-port", "-e",
+                        help="Do not bind to service ports by default. If "
+                        "set, only bind to service ports if label "
+                        "HAPROXY_{n}_ENABLE_SERVICE_PORT is set.",
+                        default=False,
+                        action="store_true")
+
+
+    parser.add_argument('--feature', dest='feature', action='store_true')
+    parser.add_argument('--no-feature', dest='feature', action='store_false')
```
scalp42: I believe these are debugging bits but could be wrong :innocent:

discordianfish:
@brndnmtthws The idea was not to break the existing behavior if people use the script directly. Feel free to change it (remember, I left Mesosphere so you need to pick that up)

lloesche:
@discordianfish @scalp42 that explanation of the group option is not correct.
So first `HAPROXY_{n}_...` the `{n}` is the port index of the app that this option applies to.
The `--group` option determines which `HAPROXY_GROUP` the script reacts to.

Example of a service that we're running on our systems:
```
HAPROXY_GROUP=external
HAPROXY_0_PORT=80
HAPROXY_0_MODE=http
HAPROXY_0_BIND_ADDR=123.45.119.105
HAPROXY_1_PORT=443
HAPROXY_1_MODE=tcp
HAPROXY_1_BIND_ADDR=123.45.119.105
```

With crontab:
```
* * * * * cd /opt/servicerouter && /opt/servicerouter/marathon_lb.py -m http://srv2:8080 -m http://srv3:8080 -m http://srv4:8080 --group external > /dev/null 2>&1
```

In this generated haproxy config the only services you'll find are those of apps with the label HAPROXY_GROUP=external. In contrast to the default behaviour of generating config for all apps.

lloesche:
@brndnmtthws I think what you're describing is what the `--group` option already does If you start the script with `--group internal` it will only generate config for apps that have the `HAPROXY_GROUP=internal` label set.

brndnmtthws:
Going to close this for now. Feel free to re-open if you think we still need it.

PR #14, opened 2015-11-18
Rename service router to load balancer or marathon-lb

aladagemre:
I think renaming was forgotten.


brndnmtthws:
Looks good to me. Any comments @discordianfish?

PR #18, opened 2015-11-19
Update documentation, make groups mandatory.

brndnmtthws:
With this, LB exposure becomes 'opt-in', rather than on-always (as with
the current implementation). To opt in to LB exposure, add a label to
the app definition to specify the LB.

For example, if you start one LB with `--group external`, and another with
`--group internal`, all apps which have the following label would be
exposed on both:

```json
{
  "labels":{
    "HAPROXY_GROUP":"external,internal"
  }
}
```
cc @discordianfish

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM debian
 
 ENTRYPOINT [ "/marathon-lb/run" ]
-CMD        [ "sse", "-m", "http://leader.mesos:8080", "--health-check" ]
+CMD        [ "sse", "-m", "http://leader.mesos:8080", "--health-check", "--group", "external" ]
 EXPOSE     80 443 8080
 
 RUN apt-get update && apt-get install -y python python-pip haproxy openssl runit \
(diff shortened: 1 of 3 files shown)


brndnmtthws:
Actually, I see now how the groups feature works. I'll simply change it to be opt-in by default.

felixb:
Am I right, that I need to put the groups label into every app now? 
Until now, everything works just out of the box. 
This change forces us to specify a artificial group into every marathon.json.

brndnmtthws:
Add `--group *` to your LB config if you want the old behaviour.

scalp42:
@brndnmtthws lets say I have two Marathon LBs with `--group internal` and `--group external`.

Within a Marathon `app.json`, what does `HAPROXY_{n}_STICKY` refers to ?

If you wanted to have the Haproxy stickiness bits for the `external` LB, would you need to specify the label as `HAPROXY_external_STICKY` ?

brndnmtthws:
Good question! the `{n}` refers to the service port index, starting from 0. It's not properly documented, so I'll update the docs accordingly.

scalp42:
Oh boy, now it makes way more sense. Thanks @brndnmtthws !

PR #24, opened 2015-11-20
Switch default appid port to 9091.

brndnmtthws:
Using port 81 by default is difficult because it's a privileged port.
Also added it back on by default, now that we're not binding to a
privileged port.

cc @felixb @scalp42


scalp42:
Thanks for heads up! We're defaulting to using static service ports for now, as good as vhosts look on paper, the issue is that you need to create DNS records to go along the task.

brndnmtthws:
Just FYI, this should not affect vhosts. It only affects things which use the magic `X-Marathon-App-Id` header.

scalp42:
Oh right, misread, carry on :innocent:

PR #29, opened 2015-12-03
Close file descriptors when forking subprocess to reload haproxy

cgbaker:
References #28

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -775,7 +775,7 @@ def reloadConfig():
     if reloadCommand:
         logger.info("reloading using %s", " ".join(reloadCommand))
         try:
-            subprocess.check_call(reloadCommand)
+            subprocess.check_call(reloadCommand, close_fds=True)
         except OSError as ex:
             logger.error("unable to reload config using command %s",
                          " ".join(reloadCommand))


brndnmtthws:
Makes sense, thanks!

brndnmtthws:
Hmm, upon closer inspection, I see that `close_fds=True` is the default in Python 3: https://docs.python.org/3/library/subprocess.html#popen-constructor

However, `close_fds=False` is the default in Python 2: https://docs.python.org/2/library/subprocess.html#popen-constructor

Since `marathon-lb.py` is now using Python 3, I don't think we need this change. Thoughts?

cgbaker:
Excellent point; I didn't notice that that the interpreter hint had changed to python3. Does the script still work with python2? If so, an explicit invocation, a la
`python3 servicerouter.py`
would still trigger the effect addressed in #28. 

If it's not necessary for marathon-lb, then it seems we can close it.

I had thought (mistakenly) that the servicerouter utility was being moved out of the main marathon repo (it's not in `master` anymore), but I see that it's still part of the distribution in the 0.12 and 0.13 releases, where it's still coding against python2. Should I refile this over in the marathon project and get these changes integrated there? I'm not sure what the long-term plan is for this utility.

brndnmtthws:
This is indeed the successor to `servicerouter.py`. It was only recently removed from the repo, however, so it probably won't be absent until the next release.

cgbaker:
So, as the current version of servicerouter doesn't work with python 2 (because of the urllib.parse requirement) and the proposed change isn't necessary for python3, I agree that this can be closed. I'll go ahead and close it.

PR #33, opened 2015-12-21
Fix the error that were abnormally terminated

fpgeek:
Fix the error that were abnormally terminated when running without the path of the healthchecks of the marathon v0.13.0


brndnmtthws:
Makes sense, thanks!

Issue #37, opened 2015-12-24
Callbacks not working


scalp42:
If you try to set up a callback url, the Marathon responses are not read correctly:

```
Traceback (most recent call last):
  File "/usr/lib/python3.4/wsgiref/handlers.py", line 137, in run
    self.result = application(self.environ, self.start_response)
  File "marathon_lb.py", line 1061, in wsgi_app
    processor.handle_event(json.loads(data))
  File "/usr/lib/python3.4/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'
10.92.42.25 - - [24/Dec/2015 00:52:41] "POST / HTTP/1.1" 500 59
```

Appears to be a Python3 issue I'm guessing per: 

- http://stackoverflow.com/questions/6862770/python-3-let-json-object-accept-bytes-or-let-urlopen-output-strings
- https://github.com/boto/boto/pull/2778

Looks like `json.loads` expect a string but we're passing a `bytes` type.

but could be wrong as always :smile:

scalp42:
Got it somewhat working by decoding and re-encoding:

```
        processor.handle_event(json.loads(data.decode('utf-8')))
        start_response('200 OK', [('Content-Type', 'text/html')])

        return "Got it\n".encode('utf-8')
```

But it's spamming a ton of traceback (the haproxy reload is working though).

PR #38, opened 2015-12-24
fix wsgi by returning bytes

scalp42:
Related to https://github.com/mesosphere/marathon-lb/issues/37

Thanks for looking at it!

On side note we should switch to something else like bottle but my python-fu might be limited.


brndnmtthws:
Seems reasonable to me. Thanks!

PR #40, opened 2016-01-04
Fix to stop interpolation of escaped asterisk

suhasgaddam:
I noticed that I cannot pass an asterisk to `--group`(using docker run or marathon-lb/run) and get the marathon-lb.py script to find all marathon apps.

This is a possible fix. Other options are `set -f` in marathon-lb/run or another `--all-groups` option in marathon-lb.py.


brndnmtthws:
Makes sense. Thanks!

PR #45, opened 2016-01-15
Fixing portIndex KeyError. when using COMMAND as protocol, there's no…

aladagemre:
Fixing portIndex KeyError.  When using COMMAND as protocol, there's no portIndex provided. This causes KeyError to raise on if check['portIndex'] == portIndex line.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -658,7 +658,7 @@ def config(apps, groups, bind_http_https, ssl_certs, templater):
             if health_check_options:
                 backends += health_check_options.format(
                     healthCheck=app.healthCheck,
-                    healthCheckPortIndex=app.healthCheck['portIndex'],
+                    healthCheckPortIndex=app.healthCheck.get('portIndex'),
                     healthCheckProtocol=app.healthCheck['protocol'],
                     healthCheckPath=app.healthCheck.get('path', '/'),
                     healthCheckTimeoutSeconds=app.healthCheck[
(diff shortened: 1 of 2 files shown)


brndnmtthws:
Looks good. Could you write a unit test for with/without the command check?

aladagemre:
I've written a test. When protocol is command, it assumes it's mode=TCP. This causes removal of some lines like below marked with +. It says lines marked with + are not generated with marathon_lb but exist in my test config. If it is normal to remove these lines, I'll be removing them on the tests and push. I don't know much about the internals of HAProxy and don't have time this week, so asking here.


    AssertionError: 'glob[536 chars]ttp\n\nfrontend marathon_https_in\n  bind *:44[184 chars]cp\n' != 'glob[536 chars]ttp\n  acl app__nginx hdr(x-marathon-app-id) -[466 chars]0s\n'
      global
        daemon
        log /dev/log local0
        log /dev/log local1 notice
        maxconn 4096
        tune.ssl.default-dh-param 2048
      defaults
        log               global
        retries           3
        maxconn           2000
        timeout connect   5s
        timeout client    50s
        timeout server    50s
        option            redispatch
      listen stats
        bind 0.0.0.0:9090
        balance
        mode http
        stats enable
        monitor-uri /_haproxy_health_check

      frontend marathon_http_in
        bind *:80
        mode http

      frontend marathon_http_appid_in
        bind *:9091
        mode http
    +   acl app__nginx hdr(x-marathon-app-id) -i /nginx
    +   use_backend nginx_10000 if app__nginx

      frontend marathon_https_in
        bind *:443 ssl crt /etc/ssl/mesosphere.com.pem
        mode http

      frontend nginx_10000
        bind *:10000
    -   mode tcp
    ?         ^
    +   mode http
    ?        + ^
        use_backend nginx_10000

      backend nginx_10000
        balance roundrobin
    -   mode tcp
    ?         ^
    +   mode http
    ?        + ^
    +   option forwardfor
    +   http-request set-header X-Forwarded-Port %[dst_port]
    +   http-request add-header X-Forwarded-Proto https if { ssl_fc }
    +   option  httpchk GET /
    +   timeout check 10s

brndnmtthws:
Yeah, that looks correct to me. Thanks for checking, and thanks for writing the test!

PR #48, opened 2016-01-29
Add vhost creation test

davidr:
In preparation for some other work, this adds a test for config generation for http vhost generation. In a newly cloned copy of the repo, the patch applies cleanly and the test succeeds.


brndnmtthws:
Wonderful, thanks!

PR #49, opened 2016-01-29
Refactor vhost ACL generation

davidr:
This is dependent on #48 

Git message:

This patch adds no new functionality, but refactors some code out of the config() function:

* Create a new function, generateHttpVhostAcl() which returns a string to add to config()'s http{,s}_frontends strings
* The code in config() is replaced by appending the return strings of generateHttpVhostAcl() to http{,s}_frontends.

This particular change likely doesn't make much sense on its own, but it required for the next pull request I'm about to submit.

For testing, I manually verified that the config generated was identical, and the tests all succeed, particularly the test introduced in #48, which tests this exact functionality.


brndnmtthws:
Again, looks great. Thanks!

PR #58, opened 2016-02-10
Fix unbound variable

occ:
With set -u, referencing $HAPROXY_SYSCTL_PARAMS will cause errors like "/marathon-lb/run: line 70: HAPROXY_SYSCTL_PARAMS: unbound variable". This effectively break existing setups.

### DIFF
--- a/run
+++ b/run
@@ -67,7 +67,7 @@ case "$MODE" in
     ;;
 esac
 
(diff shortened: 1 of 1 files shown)


brndnmtthws:
Nice catch. Thanks.

PR #63, opened 2016-02-11
In SSE mode, on failure cycle through hosts instead of hard coded first one

lloesche:
marathon-lb has always supported multiple --marathon/-m arguments. However the current code only uses those when polling. When using SSE the first host is hard coded.

This is okayish for DCOS where we usually only use one RRDNS host `-m http://master.mesos:8080` but bad for Mesos/Marathon deployments without mesos-dns. This PR fixes that by cycling through the hosts, making get_event_stream() return the next in the list whenever it's called.


brndnmtthws:
:+1: from me.

PR #67, opened 2016-02-13
Support the port parameter of health checks

fpgeek:
Marathon has allowed the port parameter to be used for health checks since v0.14.0. 
So I have added it. :)


brndnmtthws:
Makes sense, thank you!

PR #77, opened 2016-02-23
remove access to undeclared event.data variable in do_reset

flosell:
`do_reset` doesn't have an `event`, leading to crashing event processing threads.


brndnmtthws:
Oh dear, thanks!

PR #78, opened 2016-02-23
Set default secure ssl ciphers

lloesche:
This PR sets secure SSL ciphers by default, turns off SSLv3 and forces TLS v1.2 which has been default in all popular browsers since Feb 2014.

It also adds the ability to set cert options in addition to defining a certificate.
I also renamed sslCertOptions in the HAPROXY_FRONTEND_HEAD template to sslCert since the corresponding label was always called `HAPROXY_0_SSL_CERT`.

I added HAPROXY_0_BIND_OPTIONS for specifying additional bind options.
This allows users to explicitly specify insecure ciphers if they so desire for e.g. backwards compatibility.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -79,6 +79,8 @@ class ConfigTemplater(object):
       log /dev/log local1 notice
       maxconn 10000
       tune.ssl.default-dh-param 2048
+      ssl-default-bind-options no-sslv3 no-tls-tickets force-tlsv12
+      ssl-default-bind-ciphers AES128+EECDH:AES128+EDH
       server-state-file global
       server-state-base /var/state/haproxy/
       lua-load /marathon-lb/getpids.lua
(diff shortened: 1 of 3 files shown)


marathon_lb.py
```diff
@@ -633,7 +641,9 @@ def config(apps, groups, bind_http_https, ssl_certs, templater):
             backend=backend,
             servicePort=app.servicePort,
             mode=app.mode,
-            sslCertOptions=' ssl crt ' + app.sslCert if app.sslCert else ''
+            sslCert=' ssl crt ' + app.sslCert if app.sslCert else '',
+            sslCertOptions=' ' + app.sslCertOptions if app.sslCertOptions else
+            ''
```
lloesche: pep8 check complained when it was all on one line. Don't know where to best break it though, this looks ugly to me, else on the next line didn't look better.

brndnmtthws:
lgtm :+1:

lloesche:
@brndnmtthws please verify again - I just noticed while adding documentation to the README that the name sslCertOptions doesn't make any sense. They are not certificate options. If anything they are ssl options. But then sslOptions wouldn't make much sense either because to haproxy really they are bind options. So I renamed them to bindOptions and the corresponding label HAPROXY_{n}_BIND_OPTIONS. It's more generic and in addition to specifying things like ciphers and encryption methods you can also use it for anything listed in https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#5.1

brndnmtthws:
Makes sense. :+1: from me.

PR #79, opened 2016-02-23
Set more sane HAProxy defaults.

brndnmtthws:
Make sure we set 'timeout tunnel' so that websockets work correctly.


lloesche:
LGTM :+1: 
I wonder if `option http-server-close` would also make sense as a global default.

brndnmtthws:
Yes! Good thinking. Updated accordingly.

PR #82, opened 2016-02-23
Don't bind to port 80 twice when binding http listeners and redirecting http to https

tobyclemson:
Currently, when an application specifies HAPROXY_{0}_REDIRECT_TO_HTTPS, an additional 'bind *:80' statement is added to the service specific frontend along with a redirect to the HTTPS backend. This causes a conflict with the default HTTP frontend which also includes 'bind *:80'.

The correct way to handle this in haproxy is to use ACLs on a single frontend per address:port. The added test demonstrates the use case.

Unfortunately, this is a breaking change since it requires changes to the way templates are specified by consuming applications as well as the naming for custom global templates.

I couldn't think of a way of avoiding the change but if there is a better way, I can update the PR.


lloesche:
Thanks for the PR!

Doesn't this break all installations running with `--dont-bind-http-https`?
The marathon_http_in frontend is an optional default and many installations run without it.
Also note that the current implementation takes a bindAddr (`HAPROXY_0_BIND_ADDR`). `*` is just the default that's being used if no bindAddr is defined for the app.

tobyclemson:
It seems to me there are a number of cases:

**Running with --dont-bind-http-https and with no apps using HAPROXY_{0}_REDIRECT_TO_HTTPS**
The implementation here will still generate the service port frontend as usual using the same template as before

**Running with --dont-bind-http-https and with apps using HAPROXY_{0}_REDIRECT_TO_HTTPS**
The implementation will add a redirect to https to the service port frontend in the case that the request is not SSL. Since the request comes in on the service port, the redirect will also be on the service port.

**Running without --dont-bind-http-https and with no apps using HAPROXY_{0}_REDIRECT_TO_HTTPS**
The implementation is the same as before in this case

**Running without --dont-bind-http-https and with apps using HAPROXY_{0}_REDIRECT_TO_HTTPS**
The implementation will add a hostname specific redirect to the http frontend effectively redirecting to port 443 as well as the redirect for the service port frontend redirecting on the same port.

With respect to the defaults, I think all features should really be functional with the default configurations otherwise it makes it quite confusing as a user.

I can't see a way of changing the HAPROXY_HTTP_FRONTEND_HEAD, HAPROXY_HTTP_FRONTEND_APPID_HEAD and HAPROXY_HTTPS_FRONTEND_HEAD templates that would fix this problem since the HAPROXY_{0}_REDIRECT_TO_HTTPS behaviour is to always add a 'bind *:80' to the service port frontend regardless of the HEAD templates. Please let me know if I have misunderstood.

lloesche:
You can change the templates in a way that makes the catchall marathon_http_in frontend bind to one IP and your app to another. That way there's no IP conflict between the two. I see your use case for when you want incoming http requests to terminate on the marathon_http(s)_in frontends and also need the redirect.

I'm just looking for an implementation that doesn't break backwards compatibility.

The use case not covered in your examples is that when HAPROXY_{0}_REDIRECT_TO_HTTPS is set for a specific port it automatically makes the service bind to the http port as well irregardless of the configured service port. That is important because service ports in Marathon have to be globally unique. So you can't have two apps both running with service port 80. This option however allows just that.

Take the following Marathon app definition as an example:
```json
{
  "id": "/nginx",
  "cpus": 1,
  "mem": 1024,
  "instances": 2,
  "container": {
    "type": "DOCKER",
    "docker": {
      "image": "nginx",
      "network": "BRIDGE",
      "portMappings": [
        {
          "containerPort": 443,
          "hostPort": 0,
          "servicePort": 0,
          "protocol": "tcp"
        }
      ]
    }
  },
  "labels": {
    "HAPROXY_0_MODE": "http",
    "HAPROXY_0_BIND_ADDR": "10.0.0.11",
    "HAPROXY_0_PORT": "443",
    "HAPROXY_0_REDIRECT_TO_HTTPS: "true"
  }
}
```

Files:
templates/HAPROXY_HTTP_FRONTEND_HEAD
```
frontend marathon_http_in
  bind 10.0.0.10:80
  mode http
```

templates/HAPROXY_HTTP_FRONTEND_APPID_HEAD
```
frontend marathon_http_appid_in
  bind 10.0.0.10:9091
  mode http
```

templates/HAPROXY_HTTPS_FRONTEND_HEAD
```
frontend marathon_https_in
  bind 10.0.0.10:443 ssl {sslCerts}
  mode http
```

This app will be available on 10.0.0.11:80 where it redirects and 10.0.0.11:443. The PR would break that behaviour.

brndnmtthws:
Could we just change it so the new config looks like this (if the redirect flag is set):

```
frontend marathon_http_in
  bind *:80
  mode http
  acl host_test_example_com hdr(host) -i test.example.com
  acl host_test_example_com hdr(host) -i test
  redirect scheme https if !{{ ssl_fc }} host_test_example_com

frontend marathon_https_in
  bind *:443 ssl crt /etc/ssl/mesosphere.com.pem
  mode http
  use_backend nginx_10000 if { ssl_fc_sni test.example.com }
  use_backend nginx_10000 if { ssl_fc_sni test }
```

lloesche:
@brndnmtthws so the thing is, HAPROXY_{0}_REDIRECT_TO_HTTPS does not require a VHOST to be set. The marathon_http_in frontend however does. Also when I set a BIND_ADDR I potentially (always?) want the redirect to happen on that IP, not on the marathon_http_in frontend.

Maybe we could add some additional logic where if no bind_addr is set but a vhost is set and redirect_to_https is also set then add an acl to the marathon_http_in frontend otherwise use the current behaviour.

brndnmtthws:
Yeah, that seems reasonable. The primary use case I can think of is where I have some service A with vhost V which should always redirect to SSL.

brndnmtthws:
#106 is an alternative approach to this.

PR #88, opened 2016-02-26
Use TCP health check of HAProxy if app's health check protocol is "TCP"

fpgeek:
I think the TCP health check of HAProxy should be used if application's mode is "http" but health check is "tcp".
Because an application depends on it's health check.
How do you think about it?

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -756,7 +756,7 @@ def config(apps, groups, bind_http_https, ssl_certs, templater):
 
         if app.healthCheck:
             health_check_options = None
-            if app.mode == 'tcp':
+            if app.mode == 'tcp' or app.healthCheck['protocol'] == 'TCP':
                 health_check_options = templater \
                     .haproxy_backend_tcp_healthcheck_options(app)
             elif app.mode == 'http':
@@ -805,7 +805,7 @@ def config(apps, groups, bind_http_https, ssl_certs, templater):
             healthCheckOptions = None
             if app.healthCheck:
                 server_health_check_options = None
-                if app.mode == 'tcp':
+                if app.mode == 'tcp' or app.healthCheck['protocol'] == 'TCP':
                     server_health_check_options = templater \
                         .haproxy_backend_server_tcp_healthcheck_options(app)
                 elif app.mode == 'http':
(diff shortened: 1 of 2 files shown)


marathon_lb.py
```diff
@@ -805,7 +808,7 @@ def config(apps, groups, bind_http_https, ssl_certs, templater):
             healthCheckOptions = None
             if app.healthCheck:
                 server_health_check_options = None
-                if app.mode == 'tcp':
+                if app.mode == 'tcp' or app.healthCheck['protocol'] == 'TCP':
```
brndnmtthws: Maybe do a `.lower()` on these strings before the comparison.

marathon_lb.py
```diff
@@ -756,7 +759,7 @@ def config(apps, groups, bind_http_https, ssl_certs, templater):
 
         if app.healthCheck:
             health_check_options = None
-            if app.mode == 'tcp':
+            if app.mode == 'tcp' or app.healthCheck['protocol'] == 'TCP':
```
brndnmtthws: `.lower()` here too.

tests/test_marathon_lb.py
```diff
@@ -220,7 +220,7 @@ def test_config_healthcheck_command(self):
 backend nginx_10000
   balance roundrobin
   mode tcp
-  server 1_1_1_1_1024 1.1.1.1:1024
+  server 1_1_1_1_1024 1.1.1.1:1024 check inter 2s fall 11
```
brndnmtthws: Hmm, why did this change?

brndnmtthws:
Seems reasonable. Should be good to go once you address the comments above.

marathon_lb.py
fpgeek: Health check protocol of Marathon allows only "TCP". "tcp" is not allowed.

marathon_lb.py
fpgeek: "tcp" is not allowed.

tests/test_marathon_lb.py
fpgeek: Default value of "HAPROXY_BACKEND_SERVER_TCP_HEALTHCHECK_OPTIONS" has changed.

brndnmtthws: Makes sense, I missed that.

marathon_lb.py
brndnmtthws: I think you're right...although it's confusing that Marathon uses lowercase for the portmappings, but uppercase for the health checks.

brndnmtthws:
Thanks!

PR #92, opened 2016-02-26
added support for haproxy group per service port.

ajays20078:
Added support for haproxy group per service port. Currently both service port had to be part of same hagroup group. Now similar to Vhost, we can have different haproxy group per service port. This allows us to have external and internal services running on same box on different service ports, as they can be part of different haproxy group. This supports,

HAPROXY_0_GROUP
HAPROXY_1_GROUP 
and maps them to corresponding service ports.

It is backward compatible as well, so it supports HAPROXY_GROUP.

If both HAPROXY_0_GROUP and HAPROXY_GROUP are specified, the more specific definition takes preference and the service group will be added to HAPROXY_0_GROUP and not HAPROXY_GROUP.

If there are 2 services HAPROXY_0_PORT and HAPROXY_1_PORT and if HAPROXY_0_GROUP and HAPROXY_GROUP labels are defined. HAPROXY_0_PORT service is added to HAPROXY_0_GROUP and HAPROXY_1_PORT is added to HAPROXY_GROUP since HAPROXY_1_GROUP is not specified for it.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -455,6 +455,7 @@ def set_balance(x, k, v):
 def set_label(x, k, v):
     x.labels[k] = v
 
+
 def set_group(x, k, v):
     x.haproxy_groups = v.split(',')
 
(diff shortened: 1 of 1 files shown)


ajays20078:
Stashed everything to one commit and submitted a new pull request.

PR #96, opened 2016-03-01
added support for haproxy groups  per service group along with test cases and documentation

ajays20078:
added support for haproxy groups  per service group along with test cases and documentation


lloesche:
LGTM :+1:

Issue #99, opened 2016-03-03
Different path routing on same multiple host


rvalkenaers:
Is it not possible to use 2 backends on 1 virtual host and do the routing based on path

For example:
app1:
HAPROXY_VHOST = app.example.com
HAPROXY_PATH = /app1
app2:
HAPROXY_VHOST = app.example.com
HAPROXY_PATH = /app2

Both acl names in the haproxy.cfg will be called path_app_example.com so the first one always get selected.

Both requests app.example.com/app1 and app.example.com/app2 go to app1

Possible solution: include marathon app id in acl name.

brndnmtthws:
Good catch. Created a PR to address this.

rvalkenaers:
Thanks for the fix. It's working great for http.

Is it also possible to do this for https_in frontend. I think it's already possible but I don't seem to get it working.

Do I have to de someting with HAPROXY_HTTPS_FRONTEND_ACL_WITH_PATH?

cecemel:
+1

brndnmtthws:
Looks like another bug. Should be fixed with #102.

PR #102, opened 2016-03-04
Fix HTTPS path routing where only 1 vhost exists.

brndnmtthws:
To address #99.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -980,7 +980,6 @@ def generateHttpVhostAcl(templater, app, backend):
                 staging_https_frontends += https_frontend_acl.format(
                     cleanedUpHostname=acl_name,
                     hostname=vhost_hostname,
-                    path=app.path,
                     appId=app.appId,
                     backend=backend
                 )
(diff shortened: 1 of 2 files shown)


rvalkenaers:
I tested this with following config:

```
HAPROXY_1_VHOST = app.example.com
HAPROXY_1_PATH = /app1
```

But I'm getting on error on the validation of the haproxy.cfg:

`[ALERT] 066/133854 (21) : parsing [/tmp/tmp08iije8k:59] : error detected while parsing switching rule : no such ACL : 'path_app1_10016'.`


The config generated in /tmp dir looks like this:
```
frontend marathon_http_in
  bind *:80
  mode http
  acl host_app_example_com hdr(host) -i app.example.com
  acl path_app1_10016 path_beg /app1
  use_backend app1_10016 if host_app_example_com path_app1_10016

frontend marathon_https_in
  bind *:443 ssl crt /etc/ssl/example.com.pem
  mode http
  use_backend app1_10016 if { ssl_fc_sni app.example.com } path_app1_10016
```

I think the following line needs to be added to the https frontend:
`acl path_app1_10016 path_beg /app1`

so the config looks like this:

```
frontend marathon_https_in
  bind *:443 ssl crt /etc/ssl/example.com.pem
  mode http
  acl path_app1_10016 path_beg /app1
  use_backend app1_10016 if { ssl_fc_sni app.example.com } path_app1_10016
```

brndnmtthws:
Yes, you're right. The ACL is missing. Thanks! I'll update the PR in a moment.

brndnmtthws:
Give it a shot now.

rvalkenaers:
Seems to be working now.

Thanks!

PR #109, opened 2016-03-10
Peridiocally (every 5m) refresh app state.

brndnmtthws:
In case we missed one or more SSE messages, periodically refresh the app
state.


marathon_lb.py
```diff
@@ -1330,7 +1330,8 @@ def do_reset(self):
                     logger.info('stopping event processor thread')
                     return
                 if not self.__pending_reset:
-                    self.__condition.wait()
+                    if not self.__condition.wait(60):
```
brndnmtthws: I thought about making this configurable, but there are already a million config params so I'd rather not.

PR #110, opened 2016-03-11
Apply backoff on SSE stream failures.

brndnmtthws:
To prevent a thundering herd problem, apply a backoff before
reconnecting to the stream after an exception.


cmaloney:
:ship:

PR #112, opened 2016-03-14
Need 'stats socket /var/run/haproxy/socket' for seamless reloading

fpgeek:
It's referred to "http://blog.haproxy.com/2015/10/14/whats-new-in-haproxy-1-6/"


brndnmtthws:
Hmm, I thought we already had this (given https://github.com/mesosphere/marathon-lb/blob/master/service/haproxy/run#L27). I guess it got dropped.

Thanks :)

PR #120, opened 2016-03-27
Implement Mozilla's recommended cipher suites and TLS/SSL versions

pdutourgeerling:
Compatible with: Firefox 1, Chrome 1, IE 7, Opera 5, Safari 1, Windows XP IE8, Android 2.3, Java 7


pdutourgeerling:
If you agree with this change, please help me with the 79 character per line limit because I never programmed in Python.

brndnmtthws:
You can use a backslash (`\`) to escape the newline.

Can you cite the source for the recommendation?

brndnmtthws:
I adapted this to #121, and also included HSTS (https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security).

PR #122, opened 2016-03-29
IP-Per-Container Support

djosborne:
This PR enables marathon-lb to create an haproxy.cfg which routes to ip-per-container IPs instead of the agent-ip/port.

The container IP is extracted from the `ipAddresses` field returned by Marathon. However this field will be populated with the Docker Bridge IP (if the user launched the task on the docker bridge), so to indicate to mesosphere-lb whether or not it should use that value (aka if it is routable), I've introduced the `HAPROXY_IP_PER_CONTAINER` label for tasks. If set, the task IP is used. If not (default), its ignored, and the agent IP is used. This could presumably be improved with better support for user-defined docker networks.

I'll provide a simple repro ASAP. For now, just looking for comments on what design changes may be needed to get this merged.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -442,6 +442,8 @@ def set_path(x, k, v):
 def set_sticky(x, k, v):
     x.sticky = string_to_bool(v)
 
+def set_ip_per_container(x, k, v):
+    x.ip_per_container = string_to_bool(v)
 
 def set_redirect_http_to_https(x, k, v):
     x.redirectHttpToHttps = string_to_bool(v)
@@ -505,6 +507,7 @@ label_keys = {
     'HAPROXY_{0}_BACKEND_SERVER_TCP_HEALTHCHECK_OPTIONS': set_label,
     'HAPROXY_{0}_BACKEND_SERVER_HTTP_HEALTHCHECK_OPTIONS': set_label,
     'HAPROXY_{0}_BACKEND_SERVER_OPTIONS': set_label,
+    'HAPROXY_{0}_IP_PER_CONTAINER': set_ip_per_container,
 }
 
 logger = logging.getLogger('marathon_lb')
@@ -512,9 +515,11 @@ logger = logging.getLogger('marathon_lb')
 
 class MarathonBackend(object):
 
-    def __init__(self, host, port, draining):
+    def __init__(self, host, port, container_ip, container_port, draining):
         self.host = host
         self.port = port
+        self.container_ip = container_ip
+        self.container_port = container_port
         self.draining = draining
 
     def __hash__(self):
@@ -526,7 +531,7 @@ class MarathonBackend(object):
 
 class MarathonService(object):
 
-    def __init__(self, appId, servicePort, healthCheck):
+    def __init__(self, appId, servicePort, containerPort, healthCheck):
         self.appId = appId
         self.servicePort = servicePort
         self.backends = set()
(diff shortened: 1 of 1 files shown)


marathon_lb.py
```diff
@@ -1323,8 +1331,20 @@ def get_apps(marathon):
...
+                    container_port = port_mapping['containerPort']
+                    break
+            if not container_port:
+               raise Exception("No matching container port found for service port %s" % servicePort) 
+
             service = MarathonService(
-                        appId, servicePort, get_health_check(app, i))
+                        appId, servicePort, container_port, get_health_check(app, i))
+
+            if 'HAPROXY_IP_PER_CONTAINER' in marathon_app.app['labels']:
```
brndnmtthws: Perhaps use the same pattern that's used with `HAPROXY_{0}_REDIRECT_TO_HTTPS` and `HAPROXY_{0}_USE_HSTS`.

marathon_lb.py
```diff
@@ -1365,12 +1385,16 @@ def get_apps(marathon):
 
             for i in range(number_of_defined_ports):
                 task_port = task_ports[i]
+                # Arbitrarily use the container's first specified IP
+                container_ip = task['ipAddresses'][0]['ipAddress']
```
brndnmtthws: What happens where there's more than one IP?

marathon_lb.py
```diff
@@ -512,9 +512,11 @@ def set_group(x, k, v):
 
 class MarathonBackend(object):
 
-    def __init__(self, host, port, draining):
+    def __init__(self, host, port, container_ip, container_port, draining):
```
matthewdupre: These should be mixedCase.

brndnmtthws:
Code is looking better. Still need to update the existing unit tests, and add new tests for this feature.

djosborne:
@brndnmtthws I've been doing some design work around this, and want to change the implementation. Wanted to get your approval before I do so.

My conclusion is that it was incorrect of me to be using the Marathon `portMappings` fields at all. Service discovery should come from the new `"ipAddress": { "discovery" {...}}` field.

The current behavior of the portMappings fields means that if you specify one of them (containerPort, hostPort, or servicePort) you get a random value for the other two. In IP-address-per-container land, this automatic agent->container port mapping is unneeded. But users still see a random host port in the Marathon UI if they only specify a servicePort and containerPort. One solution would be to file a bug in Marathon to be able to turn off the container->host port mapping. But option 2....

The IPAddress field introduced in Marathon-v0.14.0 allows for specifiying all IP-per-container implementation details. It already has a "ports" field for this exact scenario. As per the release notes - "If an application requires IP-per-task, then it can not request ports to be allocated in the slave. It is however still possible to describe the ports that the Application's tasks expose:"

Therefore, I think we should be getting port to forward to in the haproxy entry from `ipaddress-discovery-ports`. Unfortunately, the ipaddress-discovery-ports field does not allow you to specify a servicePort to map to on the LB. I'd argue that this is fine, that IP-per-container model more closely aligns with vhost concepts than binding to a "random" (10000+) port on the load balancer, so people would use vhosts more instead. But servicePort behavior may still be useful to some, so I think a "servicePort" field should be added to ipaddress-discovery-ports. For now, we can just leave that feature out (and only support vhosts) or just use labels until Marathon adds it.

Let me know if this makes sense, I'd like to have this ready to merge by end of day today

djosborne:
I've noticed that marathon-lb's design requires each service to bind to a service port on Marathon LB. Seems that Marathon will generate a random one if a task only specifies a vhost. However, it doesn't auto-assign a service port if the user specifies a ipAddress discoveryport. I've raised an issue here:
https://github.com/mesosphere/marathon/issues/3636

brndnmtthws:
I think, generally speaking, you want to use an explicit service port, because you don't want the mapping from service -> port to change randomly. It should be deterministic, and the easiest way to do that is to require that the user specifies it.

djosborne:
@branden I'm not sure I understand - Marathon currently provides a random service port to Mesos tasks which specify only a vhost. Why would this scenario be any different?

brndnmtthws:
It's not strictly random, no. Marathon will assign the next available service port, starting from `--local_port_min` (defaults to 10000) up to `--local_port_max` (defaults to 20000).

djosborne:
Sorry, I think I've caused a misunderstanding. By "random service port", I mean "an available service port between the local_port_min and local_port max". This is allocated when a user specifies a vhost in Marathon, but doesn't happen when the ipAddress field is also supplied (as the Marathon issue I opened states). My hope is that the Marathon team can fix that bug, and then I will use modify this PR to use the `discoveryPort` as the port on the container, and map it to the servicePort allocated by Marathon.

djosborne:
Closing for more design in #139

Issue #75, opened 2016-02-23
does it support proxypass?


fanzhiwei2015:
I have been looking for the proxy pass function for marathon-lb in internet, however don't see an example.

May I know if it's possible? and how? thanks

lloesche:
Not out of the box but haproxy (the load balancer marathon-lb uses) supports something similar. Check http://blog.haproxy.com/2014/04/28/howto-write-apache-proxypass-rules-in-haproxy/ for more info.

fanzhiwei2015:
thanks for the reply, yes, I saws this as well, but how can we make it work with marathon app automatically? I also see this marathon-lb issues , and it sounds similar to the proxypass. https://github.com/mesosphere/marathon-lb/issues/64

LeePorte:
I'm actually working on creating a PR for this at the minute.

LeePorte:
Updated PR post project restructure https://github.com/mesosphere/marathon-lb/pull/137

PR #127, opened 2016-03-31
added in new templates and labels for proxying and redirection

LeePorte:
This pull request addresses https://github.com/mesosphere/marathon-lb/issues/75 and additionally adds the ability to redirect the URL root to a subdirectory.

This is a feature that I use on our internal HAProxy for accessing monitoring endpoints rather than for public consumption.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -598,8 +598,8 @@ label_keys = {
     'HAPROXY_{0}_BACKEND_SERVER_HTTP_HEALTHCHECK_OPTIONS': set_label,
     'HAPROXY_{0}_HTTP_BACKEND_PROXYPASS': set_proxypath,
     'HAPROXY_{0}_HTTP_BACKEND_REVPROXY': set_revproxypath,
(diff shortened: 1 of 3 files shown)


brndnmtthws:
The code looks fine, but you need to fix the code style, update the existing tests, and add new tests.

brndnmtthws:
Can you rebase this against master? The definitions for the labels got changed around a bit, but it should be pretty straightforward to update.

marathon_lb.py
```diff
@@ -576,6 +626,9 @@ def __init__(self, appId, servicePort, healthCheck):
         self.servicePort = servicePort
         self.backends = set()
         self.hostname = None
+        self.proxypath = ''
```
lloesche: Any reason not to default to None and then check for `if app.proxypath:` instead of `if app.proxypath != '':`? It's the style used for the rest of the code. Same for revproxypath and redirpath.

LeePorte: No reason in particular.

LeePorte:
See https://github.com/mesosphere/marathon-lb/pull/137 for continuation

PR #130, opened 2016-04-01
Task with no ports silently fails ad stop processing any more apps

ryaminal:
Using marathon there is a chance that a task won't have ports. This fixes a silent failure by exposing the problem by logging the message using the logger instead of print and if ports don't exist we use an empty list.

Symptoms of not fixing this is that not all apps are processed because as soon as any task under any app is found and doesn't have ports it jumps out of the loop and carries on as if nothing is wrong.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.git
+tests
+.*
+Dockerfile
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,8 @@ RUN apt-get update && apt-get install -y python3 python3-pip openssl libssl-dev 
     wget build-essential libpcre3 libpcre3-dev python3-dateutil socat iptables libreadline-dev \
     && pip3 install -r /marathon-lb/requirements.txt \
     && /marathon-lb/build-haproxy.sh \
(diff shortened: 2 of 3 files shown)


timcharper:
I collaborated on this patchset. The Dockerfile changes help reduce the docker image significantly; as is, `gcc`, `make`, `g++`, et al are being packaged up in the image. In other words, `--autoremove` isn't working like the Dockerfile author thought it did. Doing a separate `apt-get autoremove` call after the `apt-get remove` process resolves the issue and significantly reduces the size of the volume.

timcharper:
The logger patch 96b941b is important because `print` statements are being swallowed, making any exceptions completely invisible when using `docker logs`, `docker attach`, or even `docker run` without detaching. Switching to use the logger makes the exceptions visible.

We were encountering an exception and it was completely invisible, leaving us to do (unfortunately) needlessly debug why it wasn't working in our production cluster for hours, but was working in our dev environments. Had the exception been visible, it would've reduced the amount of time diagnosing significantly.

brndnmtthws:
Cut the image size by 35%! Nice work :smile: 

screen shot 2016-03-31 at 9 09 19 pm

PR #133, opened 2016-04-01
added missing overridable templates

mikestaub:
README and code were out of sync


brndnmtthws:
Makes sense. I've been meaning to rewrite the config stuff, and include the documentation as part of the code so we can just generate that section of the README. Keeping it up-to-date is a pain.

Issue #135, opened 2016-04-04
need configurable syslog-socket parameter via enviroment variable


ekesken:
we have an unexpected value in metric for marathon_https_in frontend at prometheus:

```
sum(increase(haproxy_frontend_http_responses_total{code="5xx"}[1m])) by (frontend)
```

because syslog-socket parameter value is hard-coded as /dev/null, we can't enable logging to see further information.

If it could be configured via enviroment variable, we can apply a workaround like:

```
docker run ... -E SYSLOG_SOCKET=/dev/log -v /dev/log:/dev/log mesosphere/marathon-lb
```

PR #136, opened 2016-04-04
made syslog-socket parameter configurable via enviroment variable, du…

ekesken:
…e to #135


brndnmtthws:
This makes sense, but rather than relying on the Dockerfile, could you put the default into the script itself? i.e., check if it's unset, and set it to the default if it's not specified.

ekesken:
I did the change, if double commit is a problem, I can recreate pull request.

brndnmtthws:
That's fine, thanks!

PR #137, opened 2016-04-05
rebases and reworked against master

LeePorte:
Rebased and reworked against master 

See https://github.com/mesosphere/marathon-lb/pull/127 for history

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -72,6 +72,9 @@ class MarathonService(object):
         self.servicePort = servicePort
         self.backends = set()
         self.hostname = None
+        self.proxypath = None
+        self.revproxypath = None
+        self.redirpath = None
         self.haproxy_groups = frozenset()
         self.path = None
         self.sticky = False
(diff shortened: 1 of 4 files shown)


marathon_lb.py
```diff
@@ -72,6 +72,9 @@ def __init__(self, appId, servicePort, healthCheck):
         self.servicePort = servicePort
         self.backends = set()
         self.hostname = None
+        self.proxypath = ''
```
lloesche: Like I commented on the last PR, any reason this doesn't default to None and then checks using `if app.proxypath:` instead of `if app.proxypath = '':`? Same for the other two.

lloesche:
Thanks for the PR @LeePorte! Small tip, if no one else is working on your branch you can rebase and then squash using e.g. `git reset --soft HEAD~N; git commit -am 'a single commit'` where N is the number of commits you want to squash. You'll just have to force push the branch then. Github will pick it up correctly. Saves you from having to close/reopen multiple PRs.

marathon_lb.py
LeePorte: There is no particular reason. If you want / expect a particular style, then I would suggest that documenting it would be a good move.

lloesche: If you look like one line up at how all the other template defaults and checks work you'll see that style.

lloesche:
LGTM :+1: 
Before merging please do
```
git reset --soft HEAD~3
git commit -am 'added in new templates and labels for proxying and redirection'
git push --force
```

PR #141, opened 2016-04-06
Replaced serverName with shortHashedServerName in haproxy configs if …

stephansnyt:
…sticky options are used, so as not to leak information about hostnames/ports


brndnmtthws:
The tests are failing, and we should also add a test specifically for this feature.

Otherwise, the PR looks great!

stephansnyt:
Closing until I add a test and pass the suite.

PR #142, opened 2016-04-06
Replaced serverName with shortHashedServerName in haproxy configs if ...

stephansnyt:
…sticky options are used, so as not to leak information about hostnames/ports


brndnmtthws:
Looks good, thanks!

PR #143, opened 2016-04-06
Allow LB to work for apps with IP per task

### LINKED ISSUE #139: Marathon-LB Server Entries don't consider ip-per-container tasks
Since v0.14.0, Marathon has supported launching tasks with their own uniquely addressable  `ipAddress` [\[0\]][0] [\[1\]][1].

Mesos-dns has been updated to check for this information and generate records which take this into account [\[2\]][2]. I'm opening this issue to suggest that Marathon-lb should too.

To expand on what a "relevant record" refers to - the current marathon-lb implementation creates the following mapping for docker bridge tasks:
`<marathon-lb-ip>:<servicePort>` :arrow_right:  `<agent-ip>:<hostPort>` :arrow_right:  `<container-ip>:<containerPort>`

IP-per-container means that the container is routable directly from Marathon-lb. So we can replace the middle entry with the container directly:
`<marathon-lb-ip>:<servicePort>` :arrow_right: `<container-ip>:<discoveryPort>`

(I use docker bridge tasks as an example, but the Mesos task implementation should similarly map directly to the task's ip and port).

The implementation should mirror the one used in the linked mesos-dns PR:
- The presence of the `ipAddress` field should trigger whether this type of entry is created or not. 
- We no longer forward requests to the agent-ip, opting instead for the task's IP which is stored in `app['tasks']['ipAddresses'][0]['ipAddress']`.
    - This field is in a list, but marathon currently only supports 1 ip address per container. So we can always grab the first entry
- We no longer forward to the host port, opting instead to use the container's service port which is specified in `app['discovery']['ports']` (possibly defaulting to 80)

I'm going to close #122 in favor of this issue, as failing design it properly resulted in a fundamentally incorrect implementation. (Marathon restricts tasks which launch with ip-per-container from also specifiying portmappings.)

[0]: https://mesosphere.github.io/marathon/docs/ip-per-task.html
[1]: https://github.com/mesosphere/marathon/releases/tag/v0.14.0
[2]: https://github.com/mesosphere/mesos-dns/pull/197

robbrockbank:
-  Fix up LB to use correct IP address and port.  For IP-per-task, use actual task IP and port.
-  Workaround Marathon issue where service port is not returned for IP-per-task apps.  The
   workaround is to auto assign a local port on the LB.  Only useful when using vhost since
   the port is not exposed anywhere, so use of DNS entries is required.

This PR does not yet include any UT updates - I didn't want to make any until the function was actually reviewed.

Fixes #139 
Works around Marathon issue: https://github.com/mesosphere/marathon/issues/3636

To workaround the Marathon issue where service ports are not returned for IP-per-task applications I've implemented a ServicePortAssigner that will assign an arbitrary local port to use for a particular application.  Ports are only auto-assigned if the app is IP-per-task and the --min-serv-port-ip-per-task and --max-serv-port-ip-per-task port range is specified (to prevent clashes with service ports assigned by Marathon).  It's only useful if you are using vhost and therefore accessing apps by name since this auto-assigned port is not exposed and is not persistent upon restart.

I can put up a PR that removes all the service port shenanigans if you'd prefer, but until Marathon is fixed to return the service ports it still won't be much use.  I've tried to implement this so that when Marathon is fixes it will just start doing the right thing.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -20,7 +20,6 @@ which marathon-lb can be reached by Marathon.
 ### Command Line Usage
 """
 
-from logging.handlers import SysLogHandler
 from operator import attrgetter
 from shutil import move
 from tempfile import mkstemp
@@ -44,8 +43,8 @@ import sys
 import socket
 import time
 import dateutil.parser
-import math
 import threading
+import traceback
 import random
 import hashlib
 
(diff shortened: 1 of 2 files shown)


brndnmtthws:
Any chance you can split it into 2 separate PRs? That'll make it a bit easier to review.

robbrockbank:
Absolutely.  In fact I'm doing that right now as it happens.   I'll close this one and re-open two separate ones.

robbrockbank:
Closing this PR in preference to another.

PR #144, opened 2016-04-07
Allow LB to work for apps with IP per task

### LINKED ISSUE #139: Marathon-LB Server Entries don't consider ip-per-container tasks

robbrockbank:
-  Fix up LB to use correct IP address and port.  For IP-per-task, use actual task IP and port.

Fixes #139 

I've added support for IP per task.  There are currently no additional UTs here, but I have fixed up the existing ones.

Note that there is a marathon issue mesosphere/marathon#3636 that will prevent IP-per-tasks from working still, so this is only part of the overall fix required.  Ideally it would be great if someone could get a fix in Marathon for this - I'm a little out of my depth to make that fix.  Could you let me know if is likely to get fixed as it's blocking me.  If not I can provide another PR that works around the missing data in the marathon-lb - but the fix is only useful in certain circumstances.


.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea
 templates/
 __pycache__
+*.pyc
```
brndnmtthws: This is a nit, but small whitespace error here (no newline at EOF).

marathon_lb.py
```diff
@@ -790,6 +815,55 @@ def get_health_check(app, portIndex):
...
+def is_ip_per_task(app):
+    """
+    Return whether the application is using IP-per-task.
+    :param app:  The application to check.
+    :return:  True if using IP per task, False otherwise.
+    """
+    return app.get('ipAddress') is not None
+
+
+def get_task_ip_and_ports(app, task):
```
brndnmtthws: Could you write a test around this new function?

brndnmtthws:
Looks great overall. We just need a test around the new function, and we're good to go.

robbrockbank:
@brndnmtthws I've updated the review with some UTs to cover all branches in that method.

What's the take on mesosphere/marathon#3636 ?  Any chance someone could fix that - I'm afraid that one is a little harder for me to get my head around.

brndnmtthws:
I'm not sure about the Marathon issue at this point, but I'll check in with the Marathon team to see if we can get some resources on it.

PR #146, opened 2016-04-07
Auto-assign service ports for IP-per-container tasks when Marathon does not

robbrockbank:
The following Marathon LB fix is a workaround for a Marathon issue: https://github.com/mesosphere/marathon/issues/3636

The above issue means Marathon does not return service ports when using apps with IP-per-task.  Upshot is Marathon-LB does not work for these apps.

Ideally, I'd really like to see a fix for the above issue, but I'm offering this PR as a temporary work-around.

The change in this PR auto-assigns a service port for each host discovery port when the app is using IP-per-task.  The ports are not persistent across restart of the marathon-lb process, and the ports are not exposed so this fix is only useful when using vhost (where you don't actually care what value of service port is chosen).  The feature is activated by specifying a min/max port as additional arguments - these ports should be chosen so that they don't clash with the service ports assigned by Marathon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,17 @@ the [templates section](Longhelp.md#templates)
 You can access the HAProxy statistics via `:9090/haproxy?stats`, and you can
 retrieve the current HAProxy config from the `:9090/_haproxy_getconfig` endpoint.
 
+Marathon LB supports load balancing for applications that use the Mesos IP-per-task
+feature, whereby each task is assigned unique, accessible, IP addresses.  For these
+tasks services are directly accessible via the configured discovery ports and there
+is no host port mapping.  Note, that due to limitations with Marathon (see 
+[mesosphere/marathon#3636](https://github.com/mesosphere/marathon/issues/3636)) 
+configured service ports are not exposed to Marathon LB for IP-per-task apps.  
+For these apps, where the service ports are missing from the Marathon app data,
+Marathon LB will automatically assign port values from a configurable range.  The range
+is configured using the `--min-serv-port-ip-per-task` and `--max-serv-port-ip-per-task`
+options.
+
 ## Deployment
 The package is currently available [from the universe](https://github.com/mesosphere/universe).
 To deploy marathon-lb on the public slaves in your DCOS cluster,
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ Marathon-lb is a tool for managing HAProxy, by consuming [Marathon's](https://gi
  * DCOS integration
  * Automated Docker image builds ([mesosphere/marathon-lb](https://hub.docker.com/r/mesosphere/marathon-lb))
  * Global HAProxy templates which can be supplied at launch
+ * Supports IP-per-task integration, such as [Project Calico](https://github.com/projectcalico/calico-containers)
 
 ### Getting started
 
@@ -43,17 +44,6 @@ the [templates section](Longhelp.md#templates)
 
 You can access the HAProxy statistics via `:9090/haproxy?stats`, and you can
 retrieve the current HAProxy config from the `:9090/_haproxy_getconfig` endpoint.
-
-Marathon LB supports load balancing for applications that use the Mesos IP-per-task
-feature, whereby each task is assigned unique, accessible, IP addresses.  For these
-tasks services are directly accessible via the configured discovery ports and there
-is no host port mapping.  Note, that due to limitations with Marathon (see 
-[mesosphere/marathon#3636](https://github.com/mesosphere/marathon/issues/3636)) 
-configured service ports are not exposed to Marathon LB for IP-per-task apps.  
-For these apps, where the service ports are missing from the Marathon app data,
-Marathon LB will automatically assign port values from a configurable range.  The range
-is configured using the `--min-serv-port-ip-per-task` and `--max-serv-port-ip-per-task`
-options.
 
 ## Deployment
 The package is currently available [from the universe](https://github.com/mesosphere/universe).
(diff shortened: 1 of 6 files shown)


marathon_lb.py
```diff
@@ -275,6 +275,87 @@ def resolve_ip(host):
             return None
 
 
+class ServicePortAssigner(object):
```
brndnmtthws: Can you move this to a new file? `marathon_lb.py` is already too big.

marathon_lb.py
```diff
@@ -1164,6 +1253,12 @@ def get_arg_parser():
     parser.add_argument("--dry", "-d",
                         help="Only print configuration to console",
                         action="store_true")
+    parser.add_argument("--min-serv-port-ip-per-task",
```
brndnmtthws: How about set a default range of 10050 to 10100?

brndnmtthws:
As much as I dislike having to do this, it seems necessary until https://github.com/mesosphere/marathon/issues/3636 is resolved in some way.

Just to help me understand, can you describe what the behaviour is under the following conditions:
 * what happens when a new app is added to Marathon which has a conflicting service port? which app wins?
 * is the port assignment always deterministic? what happens when marathon-lb restarts? will all instances of marathon-lb assign the same port to the same service? it's a bit opaque from the code

Can you also write some documentation in README.md to describe the behaviour? Lastly: a unit test is needed for `ServicePortAssigner`.

Otherwise, PR looks great, and thanks for the contribution.

robbrockbank:
Yes I'm not too fond of this change either, but needs must I guess :-)

To answer your questions:
-  If a new app is added to Marathon with conflicting service port, then behavior will be non-deterministic.  It is important that the range of service ports used for the "auto-assignment" does not clash with service ports that are returned by Marathon.
-  Port assignment is no deterministic, and is not persistent upon restart.

Not great answers I know :-/   That is why I would really like to push for a Marathon fix - it's simply not possible to have two entities managing a single set of ports without significant changes, and that would be more work than the Marathon fix I suspect.  Regarding making the port deterministic - I think that would be tricky - we could hash something or other to get a preferred port, but with only a small port range there would likely be clashes and therefore a different port value would be chosen.

It would be possible to make the port selection persistent across restarts by writing out the assignments data as config.  That would certainly improve resilience.  [I was just hoping I wouldn't have to do that ;-)  I only implemented this as a stop-gap measure].

WDYT - should I make changes you requested and perhaps add persistentence to the config?

brndnmtthws:
How about for the port assignment you use a hash on the app name? You'll have to check for duplicates and handle it somehow (maybe adjust the hashed value until there's an available slot?).

robbrockbank:
@brndnmtthws What's the importance of having the port assignment deterministic?  Is that so we can document that if you need to know the port number it's calculated like this?  If you only have a small range of available port numbers then you'll end up with a lot of clashes and so things start to become very non-deterministic.  Say tasks get created in different orders then that ordering could change which task is assigned which port.

I would have thought it would be much better to have persistence over restart rather than trying to get a deterministic port value (which I simply don't believe is possible).

brndnmtthws:
If you have multiple instances of marathon-lb, there'd be no guarantee that the service would be exposed on the same port on each instance. I'd like to keep marathon-lb as stateless as possible, so I'd rather not get into the business of trying to remember ports.

I am indeed on slack.

brndnmtthws:
I was thinking of something like this:

```
>>> import hashlib
>>> import struct
>>> buckets = 50
>>> appId = 'nginx'
>>> servicePort = 10050 + struct.unpack("<L", hashlib.sha1(appId).hexdigest()[:4])[0] % buckets
>>> print(servicePort)
10097
```

And if the service port is already in use, append something to the `appId` string and repeat the process until you get a port that's actually available. If all the ports are exhausted, then print an error and do nothing.

robbrockbank:
Ah!  I didn't realize the service port would need to be identical across multiple LB instances.  I fear this approach is not really viable then.

Suppose 'app B' is launched first and that claims a particular service port through the hash algorithm.
Suppose 'app A' is then launched and that attempts to claim the same service port, but it can't so it selects a different one.

Now suppose you have another LB that is just started and when it queries Marathon for the list of apps, you get 'app A' first followed by 'app B'.  Since 'app A' is returned first that will claim the service port that has actually been claimed for 'app B' on the other LB.

The only way I could see this working is that you always recalculate service ports every time from the app list.  This would then be deterministic across all LBs, but at the expense of the service port potentially changing dynamically.

Unless I misunderstand how this works, having the service port change dynamically would be a bad thing.

Excuse my ignorance here - how come the service port needs to be the same for all LBs?

brndnmtthws:
> The only way I could see this working is that you always recalculate service ports every time from the app list. This would then be deterministic across all LBs, but at the expense of the service port potentially changing dynamically.

Indeed, if we're assigning them automatically we must do this.

It's very important that each instance of marathon-lb behaves _exactly the same_. They shouldn't have any side effects--and they shouldn't retain any state (such as which service ports are assigned). We need to guarantee that users can connect to any LB and access their service from the same port. If, for example, someone is using the `marathon-lb.marathon.mesos` Mesos-DNS address, it'll return the IP of each marathon-lb instance, so the service port must be the same across each LB group.

If we're auto-assigned the service ports, there's really no good way to handle this. That's one of the reasons why you must specify a service port in the first place.

Anyway, I think for this particular case, it's acceptable to state in the docs that automatically assigned service ports may change when you deploy new apps which also require auto-assigned ports. For HTTP-based services, we can recommend that people rely on the `X-Marathon-App-Id` endpoint on port 9091 instead.

marathon_lb.py
robbrockbank: done,

robbrockbank:
@brndnmtthws 
UTs added, code split into a utils module, updated readme and auto-generated the long help.

marathon_lb.py
```diff
@@ -1397,10 +1347,21 @@ def process_sse_events(marathon, config_file, groups,
         if args.sse and args.listening:
             arg_parser.error(
                 'cannot use --listening and --sse at the same time')
+        if bool(args.min_serv_port_ip_per_task) != \
```
brndnmtthws: Maybe also check if `max >= min`?

utils.py
```diff
@@ -0,0 +1,179 @@
...
+        return cached_ip
+    else:
+        try:
+            logger.debug("trying to resolve ip address for host %s", host)
+            ip = socket.gethostbyname(host)
+            ip_cache[host] = ip
+            return ip
+        except socket.gaierror:
+            return None
+ip_cache = dict()
```
brndnmtthws: It's occurred to me that this should be an LRU cache, otherwise it'll leak. I can fix this in a follow-up PR.

README.md
```diff
@@ -44,6 +44,17 @@ the [templates section](Longhelp.md#templates)
 You can access the HAProxy statistics via `:9090/haproxy?stats`, and you can
 retrieve the current HAProxy config from the `:9090/_haproxy_getconfig` endpoint.
 
+Marathon LB supports load balancing for applications that use the Mesos IP-per-task
```
brndnmtthws: Could you move this section to the bottom above 'Contributing'? And give it a title. Maybe 'IP-per-task support' or something.

You can also add a bullet to the feature list at the top, like "Supports IP-per-task integration, such as Project Calico".

brndnmtthws: This is a stylistic nitpick, but use 'marathon-lb' instead of 'Marathon LB'.

brndnmtthws:
Once the comments above are addressed we should be ready to go.

marathon_lb.py
robbrockbank: done

utils.py
robbrockbank: ok - i'll leave as it is.

robbrockbank:
@brndnmtthws  It should be good to go now (after a few fights with the Python 3.x UTs ;-) )

brndnmtthws:
Excellent, thanks!

PR #155, opened 2016-04-15
Optionally make sysctl setting a soft failure.

brndnmtthws:
By default, if `systcl` fails it results in a hard failure. In other
words, the main `run` script will exit and the container will fail.

By setting `HAPROXY_SYSCTL_NONSTRICT` to a non-empty string, the
`sysctl` failures will be treated as soft failures.

This was introduced to workaround in issue where Docker will mount /proc
within the container as read-only, even when the container is running in
privileged mode.

cc @lingmann @alberts

### DIFF
--- a/run
+++ b/run
@@ -46,6 +46,16 @@ if [ -n "${MESOS_SANDBOX-}" ] && [ -d "$MESOS_SANDBOX/templates" ]; then
   cp -v "$MESOS_SANDBOX/templates/"* templates/
 fi
 
+if [ -n "${HAPROXY_SYSCTL_PARAMS-}" ]; then
+  echo "setting sysctl params to: ${HAPROXY_SYSCTL_PARAMS}"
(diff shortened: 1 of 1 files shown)


brndnmtthws:
To address #156.

lingmann:
Good news! I've discovered this problem goes away with the official docker Xenial release: Docker version 1.11.0, build 4dc599... :) Do we still want to land this? I think the main scenario where it would be useful is if `sysctl -w foo=1 bar=2` fails but still sets some of the options (and that is a scenario which would even be desirable). Otherwise, couldn't a user work around these types of failures easily enough by just customizing `HAPROXY_SYSCTL_PARAMS` for their particular environment?

I've reviewed the change and I'm a :+1: on the code if others think this would be a useful knob to have available.

brndnmtthws:
Might as well :shipit:, in case somebody runs into a similar issue.

lingmann:
:ship:

PR #159, opened 2016-04-20
acme support

vixns:
Add [haproxy-acme-validation-plugin](https://github.com/janeczku/haproxy-acme-validation-plugin/)
This plugin read challenges from the `/.well-known/acme-challenge/`folder.


brndnmtthws:
I don't think this belongs in marathon-lb. Furthermore, what happens if you have multiple HAProxy instances? How do you coordinate between them?

I think this is a better implementation: https://github.com/mesosphere/letsencrypt-dcos

Though I may be biased given that I wrote it.

heww:
@brndnmtthws btw, can we have a way to support ssl certs at app level using label (eg. HAPROXY_0_CERT=xxx where xxx is the pem file content ) ?

vixns:
letsencrypt-dcos only manage a central pem, I prefer using independant pem files and `HAPROXY_0_SSL_CERT`.

If you have multiple marathon-lb instances, you need to mount `/.well-known/acme-challenge/` from a shared storage. It's fine up to 3 marathon-lb instances ( my case ), but not for large scale deployments.

Anyway, it may be required to keep acme in haproxy for supporting tls-sni-01 challenges in the future. https://www.cloudoptimizedsmb.com/articles/20160409-00/using-haproxy-to-split-letsencrypt-acme-challenges-from-regular-traffic

brndnmtthws:
In spite of that, I'm inclined to say this doesn't belong in marathon-lb. Unless there's strong demand from the community, I'd rather not add this.

@heww That sounds like something that could be added, but it's not currently possible. That could be a separate PR.

PR #169, opened 2016-04-26
Allow to specify multiple certs in Docker ENV

lloesche:
marathon_lb.py already has support for multiple certs. However the Docker container run script currently only processes a single cert. This PR adds support for multiple certs exported as
HAPROXY_SSL_CERT0 .. HAPROXY_SSL_CERT100.

I also renamed the default cert /etc/ssl/cert.pem instead of /etc/ssl/mesosphere.com.pem
and changed the short arg `-c` to the equivalent long arg `--command` since we're using long args for everything else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,8 @@ You must set `PORTS` environment variable to allow haproxy bind to this port.
 Syntax: `docker run -e PORTS=9090 mesosphere/marathon-lb sse [other args]`
 
 You can pass in your own certificates for the SSL frontend by setting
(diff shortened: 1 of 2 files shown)


lingmann:
@lloesche this looks really solid to me... nice!! :) This PR just needs some docs added to the README.

lloesche:
Docs added and commits squashed.

kartick:
🚢 🇮🇹 :shipit:

PR #170, opened 2016-04-26
Fix minor regression from #169.

brndnmtthws:
By setting `SSL_CERTS` above the if statement (for generating the
default self-signed cert), the cert will never be generated.

cc @lloesche @lingmann @kartick

### DIFF
--- a/run
+++ b/run
@@ -14,23 +14,29 @@ else
 fi
 
 # Iterate through the arguments to see if any SSL certificate is provided.
+prev=0
 for key in "$@"
 do
 
-case $key in
+case $prev in
     --ssl-certs)
-    SSL_CERTS="yes"
+      SSL_CERTS=$key
     ;;
     *)
             # unknown option
     ;;
 esac
+prev=$key
 done
 
-SSL_CERTS="/etc/ssl/cert.pem"
 if [ -n "${HAPROXY_SSL_CERT-}" ]; then
   # if provided via environment variable, use it.
   echo -e "$HAPROXY_SSL_CERT" > /etc/ssl/cert.pem
+  if [ -n "${SSL_CERTS-}" ]; then
+    SSL_CERTS+=",/etc/ssl/cert.pem"
+  else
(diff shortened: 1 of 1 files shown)


lingmann:
Thanks good catch @brndnmtthws ... we are actually overloading the use of SSL_CERTS (earlier we set it to a boolean type, later we set it to the first cert path.) See inline code comment...

kartick:
nice catch. 🐑 🇮🇹

run
```diff
@@ -40,6 +39,7 @@ elif ! [ -n "${SSL_CERTS-}" ]; then
   cat /tmp/server-cert.pem /tmp/server-key.pem > /etc/ssl/cert.pem
   rm /tmp/server-*.pem
 fi
+SSL_CERTS="/etc/ssl/cert.pem"
```
lingmann: Can we move this to the `case $key` section above? i.e. instead of `SSL_CERTS="yes"` `SSL_CERTS="/etc/ssl/cert.pem"` so that we aren't overloading the usage of the variable?

brndnmtthws: That seems more sensible.

brndnmtthws: Actually, that won't quite work. But I realized this is still broken anyway.

kartick: What wouldn't work?

brndnmtthws:
Okay, _I think_ this is actually what we want. PTAL.

brndnmtthws:
I'm going to go ahead and merge this for now, since `master` is broken. We can revisit it in a separate PR if there are any follow-up comments.

lloesche:
Ah sorry I didn't catch it because I always tested with a custom cert. I did not mean to overwrite any previous variables. I wasn't aware of any possible `--ssl-cert` arg to the run file since the Dockerfile doesn't use it. In any case I will submit a PR that reverts all changes including my original commit and simply rename the variable I'm using.

lloesche:
I just noticed another bug with the current version which is that if --ssl-cert was provided as arg to the run file we will now supply two --ssl-cert args to marathon_lb.py because $@ is added as well.

PR #172, opened 2016-04-27
Allow to specify multiple certs in Docker ENV

lloesche:
This is the same PR as https://github.com/mesosphere/marathon-lb/pull/169 with a fix for the issue reported in https://github.com/mesosphere/marathon-lb/pull/170 as well as a fix for the situation where `--ssl-certs` would be passed to `marathon_lb.py` twice.

This should now allow any combination of passing certificates either on the commandline and/or as ENV vars or not at all resulting in the self-signed cert being generated.

Also moved the loop that looks for additional certificates (`$HAPROXY_SSL_CERT0` - `$HAPROXY_SSL_CERT100`) into the if check for `$HAPROXY_SSL_CERT` since it doesn't make any sense to have a system with user provided as well as a self signed certificate. This way a user has to explicitly set `$HAPROXY_SSL_CERT` for his first certificate and only then can add additional certs using `$HAPROXY_SSL_CERT0..`

I also fixed a bug where `echo "Define $PORTS with a comma-separated list of ports to which HAProxy binds"` would try to expand `$PORTS` instead of printing it.

### DIFF
--- a/run
+++ b/run
@@ -9,35 +9,49 @@ mkdir -p $HAPROXY_SERVICE/env
 if [ -n "${PORTS-}" ]; then
   echo $PORTS > $HAPROXY_SERVICE/env/PORTS
 else
-  echo "Define $PORTS with a comma-separated list of ports to which HAProxy binds" >&2
+  echo 'Define $PORTS with a comma-separated list of ports to which HAProxy binds' >&2
   exit 1
 fi
 
-# Iterate through the arguments to see if any SSL certificate is provided.
-for key in "$@"
-do
-
-case $key in
-    --ssl-certs)
-    SSL_CERTS="yes"
-    ;;
-    *)
-            # unknown option
-    ;;
-esac
+# Find the --ssl-certs arg if one was provided,
+# get the certs and remove them and the arg from the list
+# of positional parameters so we don't duplicate them
+# further down when we pass $@ to marathon_lb.py
+declare -i ssl_certs_pos=0
+for ((i=1; i<=$#; i++)); do
+  if [ ${!i} = '--ssl-certs' ]; then
+    ssl_certs_pos=$(($i+1))
+    break
+  fi
 done
-
+if [ $ssl_certs_pos -gt 0 ]; then
+  SSL_CERTS=${!ssl_certs_pos}
+  set -- "${@:1:$(($ssl_certs_pos-2))}" "${@:$(($ssl_certs_pos+1))}"
+  [ -n "${HAPROXY_SSL_CERT-}" ] && SSL_CERTS+=",/etc/ssl/cert.pem"
+else
+  SSL_CERTS="/etc/ssl/cert.pem"
+fi
 
 if [ -n "${HAPROXY_SSL_CERT-}" ]; then
   # if provided via environment variable, use it.
-  echo -e "$HAPROXY_SSL_CERT" > /etc/ssl/mesosphere.com.pem
-elif ! [ -n "${SSL_CERTS-}" ]; then
+  echo -e "$HAPROXY_SSL_CERT" > /etc/ssl/cert.pem
+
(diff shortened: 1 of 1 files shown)


brndnmtthws:
👍 

I tested the `lloesche_multicertfix` tag on my cluster with success.

lloesche:
Nice, let's have @lingmann or @kartick look over it one last time just for a sanity check before merging.

run
```diff
@@ -9,54 +9,52 @@ mkdir -p $HAPROXY_SERVICE/env
...
-prev=$key
+# Find the --ssl-certs arg if one was provided,
+# get the certs and remove them and the arg from the list
+# of positional parameters so we don't duplicate them
+# further down when we pass $@ to marathon_lb.py
+declare -i ssl_certs_pos=0
+for ((i=1; i<=$#; i++)); do
+  if [ ${!i} = '--ssl-certs' ]; then
+    ssl_certs_pos=$(($i+1))
+    break
```
lloesche: FYI I just added this `break` and squashed the commits. Other than that it's the same code you already tested.

run
```diff
@@ -9,54 +9,52 @@ mkdir -p $HAPROXY_SERVICE/env
...
-    ;;
-esac
-prev=$key
+# Find the --ssl-certs arg if one was provided,
+# get the certs and remove them and the arg from the list
+# of positional parameters so we don't duplicate them
+# further down when we pass $@ to marathon_lb.py
+declare -i ssl_certs_pos=0
+for ((i=1; i<=$#; i++)); do
+  if [ ${!i} = '--ssl-certs' ]; then
```
kartick: Should this be `-eq` or is `=` correct?

kartick:
Can we re-write this in Python? Lot of complex stuff going on in `bash` and it will be a maintenance nightmare.

run
lloesche: -eq is for numbers, = for strings, so = is correct.

lloesche:
True though looking at marathon_lb.py our little Cabo servicerouter proof of concept hack has grown into quite the monster over the past two years. So rewriting it in Python is not a guarantee it's going to be more maintainable :smile:
The entire project could use some refactoring and architecture love.

kartick:
I would recommend opening a ticket in the backlog. If someone gets to it, good. :) Anyway, the current version looks good. Let's see what @lingmann says.

lingmann:
Ughh yeah this is getting complicated. Do we even need to support passing certs via the command line? It was just a boolean flag to begin with, so why not just keep that behavior and require the user to set custom certs via env vars? I'd prefer to keep our bash scripts super simple and I'm not sure what value this is providing.

lingmann:
@lloesche leaving this up to you if you want to merge or not since the code looks reasonable. But please take my comments into consideration about complexity and whether we really need this feature.

lloesche:
@lingmann so, it was not just a boolean flag. This is what screwed me over in the first place. What happens there is a lot of implicit intertwined things that I don't like at all but not to break backwards compatibility I implemented it the way it is now.

To explain:
The boolean just checks if `--ssl-certs` is provided as argument. If not and if no certificate was provided via the environment, then it generated the self signed certificate. If `--ssl-certs` is provided and no cert was provided via the env then that would implicitly be passed through to the Python script further down where "$@" is just completely appended to the call. But not before removing some of it's arguments.

I first thought I could just remove that entire piece of code since it's not being used by the Dockerfile but then noticed that while the ENTRYPOING is the Python script, all the args in CMD are user configurable. So it is possible that we have users of marathon-lb that start the Docker container with custom ssl certificates by volume mounting them into the container and passing `--ssl-certs` as an argument.

Took me quite a while to wrap my head around everything that's happening in those few lines of code and it really feels like a jenga tower. So I absolutely agree with @kartick that this needs refactoring. However for right now I feel that I've completely understood all the intricacies and the change should now work as expected.

PR #174, opened 2016-04-27
Refactor `process_json` and `get_app_info` in bluegreen_deploy.py

JamesAwesome:
It was previously very hard to understand what was happening in the `bluegreen_deploy.py` script with the looping constructs in `process_json` and `get_app_info`, instead refactor these methods to be more straight forward and readable.


brndnmtthws:
Nice :) Thanks!

That's what I get for not actually knowing how to write Python. Maybe I'll learn it one day.

PR #176, opened 2016-04-29
Remove TLS 1.0 from the default HAProxy SSL options

mattkirman:
PCI-DSS 3.1 states that SSL and TLS 1.0 can no longer be used for new implementations after June 30, 2016 (existing implementations now have until June 2018 to migrate).

This PR removes TLS 1.0 support from HAProxy as most browsers/OS combinations currently support TLS 1.1 or greater by default (the exception being anyone stuck on Windows Vista or earlier). As this is effectively a breaking change for those that have to support old clients it may be desirable to place this behind a flag instead?

Ref: http://blog.pcisecuritystandards.org/migrating-from-ssl-and-early-tls


brndnmtthws:
Seems reasonable. I prefer "secure by default", but obviously this will break compatibility for old clients.

I'd like to gather some feedback before I merge this.

air:
+1 for secure by default.
If customers were to ask for it (likely they won't) we could add an opt-in flag to re-enable TLS 1.0.

lloesche:
@brndnmtthws my two cents: I prefer secure by default as well. Let's create a tag and add a disclaimer in the README that tells folks something along the lines of
```
Note: For security reasons versions of marathon-lb vX.XX and later disable TLS 1.0 by default. If you require TLS 1.0 you will have to re-enable it manually in the templates.
```

PR #184, opened 2016-05-05
correct bug when scaling app instances

JamesAwesome:
Small bug fix 😦 

~~We need to compare 150% of instances when scaling against the
old_app (previously existing_app) instances. otherwise we always
scale to deployment target on first go.~~

Update: Better explanation added below...

rename scale_up_app_instances to scale_new_app_instances to avoid
confusion.

### DIFF
--- a/bluegreen_deploy.py
+++ b/bluegreen_deploy.py
@@ -307,14 +307,14 @@ def waiting_for_drained_listeners(listeners):
 
 
 def scale_new_app_instances(args, new_app, old_app):
-    """Scale the app by 150% of its existing instances until we hit the
+    """Scale the app by 150% of its existing instances until we
        meet or surpase old_app instances. At which point go right to
        the new_app deployment target
     """
     instances = (math.floor(new_app['instances'] +
                  (new_app['instances'] + 1) / 2))
     if instances >= old_app['instances']:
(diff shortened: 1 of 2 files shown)


brndnmtthws:
Hmm, can you elaborate? Could you can also write a test for the `scale_new_app_instances ` function?

JamesAwesome:
Sure thing. In the original code the instances were stepped up by 150% of the total existing deployed instances (10 instances => 15 instances). If however 15 instances was >= the amount of existing old application instances, the script instead jumped right to the deployment target.

original code:
```python
        instances = math.floor(app['instances'] + (app['instances'] + 1) / 2)
        if instances >= existing_app['instances']:
            instances = target_instances
        logger.info("Scaling new app up to {} instances".format(instances))
```

I slipped up and in the original refactor i did:
```python
        instances = math.floor(app['instances'] + (app['instances'] + 1) / 2)
        if instances >= app['instances']:  # Note i resued "app" here, not "old_app" or "existing_app"
            instances = target_instances
        logger.info("Scaling new app up to {} instances".format(instances))
```

This bugfix addresses that 😄

brndnmtthws:
Ah, makes sense. Thanks!

PR #186, opened 2016-05-06
Fix regression in ZDD draining timeout.

brndnmtthws:
If the max wait timeout is exceeded, we should just continue with the deployment.

cc @JamesAwesome

### DIFF
--- a/Longhelp.md
+++ b/Longhelp.md
@@ -848,7 +848,7 @@ Specified as `HAPROXY_{n}_BIND_OPTIONS`.
 
 Set additional bind options
 
-Ex: `HAPROXY_0_BIND_OPTIONS = 'ciphers AES128+EECDH:AES128+EDH force-tlsv12 no-tlsv10 no-sslv3'`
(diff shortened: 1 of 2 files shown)


JamesAwesome:
Would this cause an infinite Loop though? If `check_time_and_sleep` Timesout once it would in theory Timeout every time the loop restarts (since the loop restarts and uses the same timeout which is now older than `max_wait`.). If the user ran this script with `--force`  they would hit L#295 then either hit the continue on L#303 or error due to something like `drained_task_ids` not being in scope on L#302.

We could put a `break` on line L#289 or refactor `swap_bluegreen_apps` or maybe have the `while` loop do something like:

```pythoon
    while max_wait_not_exceeded(timestamp):
         sleep(args.step_delay)
         [...]
```

Then simply log a warning and stop looping once we've exceeded the max wait time.

brndnmtthws:
Hmm, okay. I'll have to take another look at this.

The behaviour I want is for the deploy to continue to the next step once we've waited N seconds (with a default of 300 or 5 minutes).

JamesAwesome:
Let me know if i can help in anyway 👍  I think something similar to what I showed above would be a good idea in that scenario.

PR #187, opened 2016-05-09
correct and add test for `swap_bluegreen_apps` timeout

JamesAwesome:
@brndnmtthws This tests that when `args.max_wait` is exceeded, the function `swap_bluegreen_apps` will log the timeout and return false instead of throwing an exception. (thus allowing the script to exit naturally) This is done by using the new method `max_wait_not_exceeded` as the condition for the `while` loop in `swap_bluegreen_apps`. This means that a timeout will only happen after a full iteration of the `while` loop, ensuring we are not left with any undeleted tasks which we know about.

As well since we no longer raise an exception on timeout I've simply set the scripts exit code based on the return value of `do_bluegreen_deploy`. This allows the script to exit gracefully on timeout but with a non-zero exit code to indicate failure.

This should prevent an infinite loop due to catching the Timeout exception and never exiting from the `while True` loop.

### DIFF
--- a/bluegreen_deploy.py
+++ b/bluegreen_deploy.py
@@ -14,6 +14,7 @@ import re
 import math
 import six.moves.urllib as urllib
 import socket
+import sys
 
 
 logger = logging.getLogger('bluegreen_deploy')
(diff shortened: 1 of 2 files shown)


brndnmtthws:
This looks better, although it's still not quite right. The deploy should proceed (rather than exiting) when it times out. I'll merge this and follow up with a separate PR when I get a chance.

PR #191, opened 2016-05-13
A ZDD should continue when draining times out.

brndnmtthws:
cc @JamesAwesome

### DIFF
--- a/bluegreen_deploy.py
+++ b/bluegreen_deploy.py
@@ -193,9 +193,13 @@ def _has_pending_requests(listener):
     return int(listener.qcur or 0) > 0 or int(listener.scur or 0) > 0
 
 
-def select_drained_listeners(listeners):
+def select_drained_listeners(listeners, timed_out):
     draining_listeners = [l for l in listeners if l.status == 'MAINT']
-    return [l for l in draining_listeners if not _has_pending_requests(l)]
+    drained_listeners = \
+        [l for l in draining_listeners if not _has_pending_requests(l)]
+    if timed_out:
+        return draining_listeners
+    return drained_listeners
 
 
 def get_svnames_from_task(task):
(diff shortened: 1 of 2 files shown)


bluegreen_deploy.py
```diff
@@ -193,9 +193,13 @@ def _has_pending_requests(listener):
...
 
 
-def select_drained_listeners(listeners):
+def select_drained_listeners(listeners, timed_out):
     draining_listeners = [l for l in listeners if l.status == 'MAINT']
-    return [l for l in draining_listeners if not _has_pending_requests(l)]
+    drained_listeners = \
+        [l for l in draining_listeners if not _has_pending_requests(l)]
+    if timed_out:
+        return draining_listeners
```
JamesAwesome: I think it might be more clear if we separated the responsibility here into two functions `select_draining_listers` and `select_drained_listeners`

```python
def select_draining_listeners(listeners):
    return [l for l in listeners if l.status == 'MAINT']

def select_drained_listeners(listeners):
    draining_listeners = select_draining_listeners(listeners)
    return [l for l in draining_listeners if not _has_pending_requests(l)]
```

We can then switch on the `timed_out` bool at a higher level. This will allow people to more easily know if they are killing drained listeners or drainING listeners and aid the readability.

eg:
```python
if not timed_out:
    listeners_to_kill = select_drained_listeners(listeners)
else:
    listeners_to_kill = select_draining_listeners(listeners)
```

JamesAwesome:
Okay, I think I get what you're going for!

Could you confirm my understand of `swap_bluegreen_apps`?:
* wait until `max_wait` for listeners to drain.
* if we hit max_wait, delete any tasks which are draining (regardless of if they have fully drained) then start over the function with a new timestamp (so each iteration of `swap_bluegreen_apps`, not the entire script waits until `max_wait`, the script never times out)

brndnmtthws:
Abandoning this in favour of #192e

Issue #190, opened 2016-05-13
bluegreen_deploy Client Error: Conflict for url


cgmartin:
Hello, I'm running a single VM with Mesos/Marathon/Marathon-lb and getting the following error during a bluegreen_deploy. Any ideas?

```
$ docker run -it -v ${PWD}:/tmp/deploy --entrypoint /bin/bash \
mesosphere/marathon-lb -c "./bluegreen_deploy.py --json /tmp/deploy/microbot-example.json \
--marathon http://10.0.0.3:8080 --marathon-lb http://10.0.0.3:9090 \
--syslog-socket /dev/null --force"

bluegreen_deploy: Final app definition:
{
  "container": {
    "docker": {
      "image": "dontrebootme/microbot:v2",
      "network": "BRIDGE",
      "portMappings": [
        {
          "containerPort": 80,
          "hostPort": 0,
          "protocol": "tcp",
          "servicePort": 8001
        }
      ]
    },
    "type": "DOCKER"
  },
  "cpus": 0.03,
  "healthChecks": [
    {
      "gracePeriodSeconds": 10,
      "intervalSeconds": 30,
      "path": "/",
      "protocol": "HTTP"
    }
  ],
  "id": "/microbot-green",
  "instances": 1,
  "labels": {
    "HAPROXY_0_PORT": "8000",
    "HAPROXY_APP_ID": "microbot",
    "HAPROXY_DEPLOYMENT_ALT_PORT": "8001",
    "HAPROXY_DEPLOYMENT_COLOUR": "green",
    "HAPROXY_DEPLOYMENT_GROUP": "microbot",
    "HAPROXY_DEPLOYMENT_STARTED_AT": "2016-05-13T16:37:31.438223",
    "HAPROXY_DEPLOYMENT_TARGET_INSTANCES": "2",
    "HAPROXY_GROUP": "external"
  },
  "mem": 64.0,
  "upgradeStrategy": {
    "maximumOverCapacity": 0.1,
    "minimumHealthCapacity": 0.8
  }
}
bluegreen_deploy: Existing app running 2 instances, new app running 1 instances
bluegreen_deploy: Found 2 app listeners across 1 HAProxy instances
bluegreen_deploy: Existing app running 2 instances, new app running 1 instances
bluegreen_deploy: Found 2 app listeners across 1 HAProxy instances
bluegreen_deploy: Existing app running 2 instances, new app running 1 instances
[...]

PR #192, opened 2016-05-14
Bugfix - Scale new_app separately from killing old_app

JamesAwesome:
@brndnmtthws this should address both #190 and #191 Normally I would just work on the original PR with you, but I saw a chance to kill two birds with one PR.

There is one difference here, and that's that i simply wait for apps to be fully drained (no qcur and no scur) before i delete their tasks.

### DIFF
--- a/bluegreen_deploy.py
+++ b/bluegreen_deploy.py
@@ -231,12 +231,9 @@ def max_wait_not_exceeded(max_wait, timestamp):
     return (time.time() - timestamp < max_wait)
 
 
-def swap_bluegreen_apps(args, new_app, old_app, timestamp):
+def find_tasks_to_kill(args, new_app, old_app, timestamp):
     while max_wait_not_exceeded(args.max_wait, timestamp):
         time.sleep(args.step_delay)
-
-        old_app = fetch_marathon_app(args, old_app['id'])
-        new_app = fetch_marathon_app(args, new_app['id'])
 
         logger.info("Existing app running {} instances, "
                     "new app running {} instances"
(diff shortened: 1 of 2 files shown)


brndnmtthws:
> There is one difference here, and that's that i simply wait for apps to be fully drained (no qcur and no scur) before i delete their tasks.

The thing is, the timeout is there to handle the case of "what happens if my backend never drains?" i.e., maybe there are very long lived connections, and you eventually (say, after 5 minutes) want to cut them off so the deploy can proceed.

Minus that detail, the PR looks good.

By the way, checking qcur and scur is kind of pointless, but it's okay to do it anyway. The only indicator that there are still connections to the draining backends is the number of HAProxy PIDs.

bluegreen_deploy.py
```diff
@@ -267,31 +264,45 @@ def swap_bluegreen_apps(args, new_app, old_app, timestamp):
         if waiting_for_drained_listeners(listeners):
             continue
 
-        drained_task_ids = \
-            find_drained_task_ids(old_app, listeners, haproxy_count)
+        return find_drained_task_ids(old_app, listeners, haproxy_count)
+
+    logger.info('Timed out waiting for drained tasks, Continuing...')
+    return []
```
brndnmtthws: As per my other comment, we should just return all the backends marked for draining at this point.

JamesAwesome:
Cool, got it! 😄 Will most likely get this PR updated Monday afternoon-ish 👍

JamesAwesome:
Okay, so the way this is setup now is that if we time out during `find_tasks_to_kill` we return task ids for any listeners which are marked 'MAINT' (regardless of scur/qcur) across all haproxy backends. However, if we find any listeners with fully drained tasks (no scur/qcur) prior to the timeout we kill those tasks first.

brndnmtthws:
This is looking right, thank you.

Issue #197, opened 2016-05-17
Make the -m and --group arg configurable for Docker image


lloesche:
Atm. one has to override the CMD to customise `--group` and `-m`. We should create a patch that makes it overridable via an ENV var using the same defaults we've currently hard coded in the Dockerfile. The patch must check if the user hasn't provided their own `--group` or `-m/--marathon` via the CMD already and if not append either the default or if defined the value from the environment. Good env names would be `HAPROXY_GROUP` and `HAPROXY_MARATHON`.

PR #199, opened 2016-05-19
Allow to specify --group and --marathon value via ENV

lloesche:
This implements #197 

Should be fully backwards compatible.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM debian:stretch
 
 ENTRYPOINT [ "/marathon-lb/run" ]
-CMD        [ "sse", "-m", "http://master.mesos:8080", "--health-check", "--group", "external" ]
+CMD        [ "sse", "--health-check" ]
 EXPOSE     80 81 443 9090
 
 COPY  . /marathon-lb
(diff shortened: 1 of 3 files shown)


brndnmtthws:
Looks good. Have you tested it?

lloesche:
I tested that when started with `HAPROXY_GROUP=myapplb` `HAPROXY_MARATHON=https://leader.mesos:8443` it results in

```
   15 ?        S+     0:00 python3 /marathon-lb/marathon_lb.py --syslog-socket /dev/null --haproxy-config /marathon-lb/haproxy.cfg --ssl-certs /etc/ssl/cert.pem --command sv reload /marathon-lb/service/haproxy --sse --marathon https://leader.mesos:8443 --group myapplb --health-check
```

and when started without any extra args it runs with the same defaults it's currently using.

brndnmtthws:
Cool, thanks!

brndnmtthws:
This doesn't seem to work. I'm getting the following error:

```
/marathon-lb/run: line 116: MARATHON_ARGS[@]: unbound variable
```

lloesche:
I'm looking into it. Did you provide any special arguments?

PR #201, opened 2016-05-20
Hotfix: unbound variable error

lloesche:
Fixes unbound variable error introduced in #199


lloesche:
Merging myself since master is broken right now.

Issue #209, opened 2016-06-01
zdd: missing task IP in get_svnames_from_task


h0tbird:
A unique, friendly name for the backend server is forged [here](https://github.com/mesosphere/marathon-lb/blob/master/marathon_lb.py#L439-L452). In my case it looks like this `worker-1_cell-1_dub_xnood_com_10_0_0_243_31096`.

Later `zdd.py` forges `svnames` [here](https://github.com/mesosphere/marathon-lb/blob/master/zdd.py#L202-L205) with missing task IP so it looks like this `worker-1_cell-1_dub_xnood_com_31096`

Therefore [this](https://github.com/mesosphere/marathon-lb/blob/master/zdd.py#L228) condition is always false.

JamesAwesome:
@brndnmtthws I think i see the problem here, could probably have a PR by tmw or friday

h0tbird:
I guess the solution is to take the IP address from the `task` structure and concatenate it to the `prefix` variable defined in `get_svnames_from_task`. But that leads me to anticipate another problem in my setup.

This is a sample `task` structure:
```yaml
{
  "appId": "/nginx-blue",
  "healthCheckResults": [
    {
      "alive": true,
      "consecutiveFailures": 0,
      "firstSuccess": "2016-06-02T09:01:39.215Z",
      "lastFailure": null,
      "lastFailureCause": null,
      "lastSuccess": "2016-06-02T09:07:10.445Z",
      "taskId": "nginx-blue.9934795e-28a0-11e6-b2be-0242f110140f"
    }
  ],
  "host": "worker-1.cell-1.dub.xnood.com",
  "id": "nginx-blue.9934795e-28a0-11e6-b2be-0242f110140f",
  "ipAddresses": [
    {
      "ipAddress": "10.128.2.162",
      "protocol": "IPv4"
    }
  ],
  "ports": [
    31096
  ],
  "slaveId": "7a619ed4-e6e2-4fee-9d8b-ccfd2bb0fca2-S1",
  "stagedAt": "2016-06-02T09:01:31.180Z",
  "startedAt": "2016-06-02T09:01:33.903Z",
  "version": "2016-06-02T09:01:31.112Z"
}
```

The IP address reported is `10.128.2.162` which belongs to my overlay network (I use flannel at the moment) whereas the IP used by `marathon-lb.py` is `10.0.0.243` which is the Mesos agent IP, so it won't match neither.

[...]

PR #211, opened 2016-06-02
Match logic for deriving haproxy svname from task in zdd.py

JamesAwesome:
This will derive a marathon task's unique backend/svname according to the same logic as the marathon_lb.py script via `utils.get_task_ip_and_ports` method: https://github.com/mesosphere/marathon-lb/blob/master/utils.py#L154 

I've also gone ahead and cleaned up some unused imports in zdd.py...

This should address Issue #209


JamesAwesome:
@h0tbird if you could test that'd be cool. Despite not using the mesos agent ip in this svname as you described, the svname we now derive should be done so using the same logic as marathon_lb.py when choosing the svname for use in haproxy. So long as the two svnames match that's all that should matter.

h0tbird:
Thank you for such a fast response, this is my current situation:

`marathon-lb.py` produces `worker-2_cell-1_dub_xnood_com_10_0_0_124_31195` (Agent IP)
`zdd.py` produces `worker-2_cell-1_dub_xnood_com_10_128_4_98_31195` (container IP)

If the same logic is used in both then I wonder if this is a matter of having enough context at  execution time. Any idea?

JamesAwesome:
Not sure. Assuming in #209 the task json you copied into the comments produced the name `worker-2_cell-1_dub_xnood_com_10_0_0_124_31195` I'm not even sure where the `10_0_0_124` would be coming from. May be a question for somebody else on the project.

brndnmtthws:
Thanks!

PR #212, opened 2016-06-02
IP-per-container workaraund

h0tbird:


h0tbird:
Ops! Sorry this should not be a PR

PR #216, opened 2016-06-07
Changed zdd to work with svnames generated by haproxy

ajays20078:
Changed zdd script to have a option to specify whether ip per container is being used, currently it assumes svnames generated by marathon-lb to be of the form [mesos_slave_ip_container_ip_port](https://github.com/mesosphere/marathon-lb/blob/master/zdd.py#L216) in certain cases, which is not true, marathon-lb is generating svnames in the form of [mesos_slave_ip_port](https://github.com/mesosphere/marathon-lb/blob/master/marathon_lb.py#L1127), so adding a flag which when set to True works with the svnames generated by marathon-lb currently.

Example: 

If mesos slave ip is 10.0.6.25 and container ip is 172.17.1.71 and host port is 31184, zdd was assuming haproxy generated a svname of the form:
**10_0_6_25_172_17_1_71_31184**

But marathon-lb is generating svname if the form **10_0_6_25_31184** in the add_backend [function call](https://github.com/mesosphere/marathon-lb/blob/master/marathon_lb.py#L1127), where it just adds task['host'] as service name. But zdd was adding [**task['host']_task['ip']_port**](https://github.com/mesosphere/marathon-lb/blob/master/zdd.py#L216) , which is not reflected in marathon-lb. 

So zdd was going into infinite loop when haproxy generates svname without container ip with the following output being printed infintely. 
**zdd: Existing app running 5 instances, new app running 5 instances
zdd: Found 10 app listeners across 1 HAProxy instances**

Added a flag --no-ip-per-container , which tells zdd that the haproxy svnames doesn't include container ip and just includes host ip. This will prevent it from going into infinite loop mentioned above. This flag by default is set to False , so that it doesn't break existing functionality.

### DIFF
--- a/tests/test_zdd.py
+++ b/tests/test_zdd.py
@@ -79,9 +79,9 @@ class TestBluegreenDeploy(unittest.TestCase):
         haproxy_instance_count = 2
         apps = json.loads(open('tests/zdd_app_blue.json').read())
         app = apps['apps'][0]
-
+        args = Arguments()
         results = \
-            zdd.find_drained_task_ids(app,
+            zdd.find_drained_task_ids(args, app,
                                       listeners,
                                       haproxy_instance_count)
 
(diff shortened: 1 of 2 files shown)


ajays20078:
Travis build failed, correcting them and stashing them into one PR , closing this.

PR #218, opened 2016-06-07
Changed zdd to work with svnames generated by haproxy

ajays20078:
Changed zdd script to have a option to specify whether ip per container is being used, currently it assumes svnames generated by marathon-lb to be of the form [mesos_slave_ip_container_ip_port](https://github.com/mesosphere/marathon-lb/blob/master/zdd.py#L216) in certain cases, which is not true, marathon-lb is generating svnames in the form of [mesos_slave_ip_port](https://github.com/mesosphere/marathon-lb/blob/master/marathon_lb.py#L1127), so adding a flag which when set to True works with the svnames generated by marathon-lb currently.

Example: 

If mesos slave ip is 10.0.6.25 and container ip is 172.17.1.71 and host port is 31184, zdd was assuming haproxy generated a svname of the form:
**10_0_6_25_172_17_1_71_31184**

But marathon-lb is generating svname  **10_0_6_25_31184** in the add_backend [function call](https://github.com/mesosphere/marathon-lb/blob/master/marathon_lb.py#L1127), where it just adds task['host'] as service name. But zdd was adding [**task['host']_task['ip']_port**](https://github.com/mesosphere/marathon-lb/blob/master/zdd.py#L216) , which is not reflected in marathon-lb. 

So zdd was going into infinite loop when haproxy generates svname without container ip with the following output being printed infintely. 
**zdd: Existing app running 5 instances, new app running 5 instances
zdd: Found 10 app listeners across 1 HAProxy instances**

Added a flag --no-ip-per-container , which tells zdd that the haproxy svnames doesn't include container ip and just includes host ip. This will prevent it from going into infinite loop mentioned above. This flag by default is set to False , so that it doesn't break existing functionality.


ajays20078:
Travis build for python 3.5 seems to have due to some unrelated issue, its passed for 2.7 and 3.4:


**3.5 is not installed; attempting download
The command "curl -s -o python-3.5.tar.bz2 ${archive_url}" failed and exited with 18 during.
Your build has been stopped.**

is the build error.

h0tbird:
How about adding the `--no-ip-per-container` flag to `marathon-lb.py`? That would cover all possible combinations. Additionally I think it is better to rename the flag to `--no-ip-per-task`.

brndnmtthws:
> How about adding the `--no-ip-per-container` flag to `marathon-lb.py`? That would cover all possible combinations. Additionally I think it is better to rename the flag to `--no-ip-per-task`.

This sounds reasonable, but let's leave the `--no-ip-per-container` flag in with a note that it's deprecated and will be removed in a later version.

ajays20078:
On a closer look, the issue is not with marathon_lb.py

- The issue is marathon_lb makes use of the **get_task_ip_and_ports** function from the [utils.py] (https://github.com/mesosphere/marathon-lb/blob/master/utils.py#L154) module to get the task ip (which takes care of checking if its ip per container or not).

- whereas zdd.py was using its own implementation of [**_get_task_ipaddress**](https://github.com/mesosphere/marathon-lb/blob/master/zdd.py#L200) , which is assuming it is ip per container all the time. 

- Add to this the misleading test cases which is not using [**ip_per_container**](https://github.com/mesosphere/marathon-lb/blob/master/tests/zdd_apps.json#L88) in the request json , but it uses ip per container in the [generated_config](https://github.com/mesosphere/marathon-lb/blob/master/tests/haproxy_stats.csv#L31)

Have modified the PR to address all these issues with zdd also using the same **get_task_ip_and_ports** functions from the utils.py and modified the test case response to reflect the absence of ip_per_container in the request.

Please review it.

brndnmtthws:
Looks okay to me.

cc @JamesAwesome

JamesAwesome:
Yeah looks good. Looks like my bug was assuming if `task[ipAddresses']` was set then ip-per-container was in use. I think using `get_task_ip_and_ports` is what I should have done in the first place. 👍  from me

JamesAwesome:
👍  from me, using `get_task_ip_and_ports` is a better way to address this.

PR #228, opened 2016-06-21
Haproxy map support to speed up vhost to backend lookup.

ajays20078:
Currently vhost to backend lookup happens in the traditional way, where each rule is scanned to lookup the corresponding backend, this can add considerable latency when the number of apps/vhosts are large in number (especially for apps/vhosts which appear at the rear end of the rules comparison).

Ex:

frontend xyz
   {other_lines}
   use_backend backend1 if { hdr(Host) -i myapp.domain1.com }
   use_backend backend2 if { hdr(Host) -i myapp.domain2.com }
   ....
   ....
   ....
   use_backend backendN if { hdr(Host) -i myapp.domainN.com }

Here the lookups for domainN, N-1, N-2 etc . can take considerable time, since these comparisons are sequential. Hence this PR supports a `--haproxy-map` , which creates a lookup map which is supported by haproxy, so for the above example when we use haproxy-map, we get

frontend xyz
  {other_lines}
  use_backend %[req.hdr(host),lower,map_dom(/path/to/map)]

and /path/to/map contains:

myapp.domain1.com backend1
myapp.domain2.com backend2
...
...
...
myapp.domainN.com backendN,

which is a hash based lookup dict, this way domainN, N-1 , N-2 etc, lookups will be faster.

Currently it just works for host header and x-marathon-app-id header, it retains the rule based comparison for bath based routing and auth.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ $ ./marathon_lb.py --marathon http://localhost:8080 --group external --skip-vali
 ```
 
 ### Using Haproxy maps for backend lookup.
-You can use haproxy maps to speed up web application (vhosts) to backend lookup. This can be very useful for large installations where vhost to backend lookup can take considerable time. This can be done by using --haproxy-map flag.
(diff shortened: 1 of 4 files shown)


ajays20078:
Few functions are not supported in python3. Ignore this PR, will fix and open a new PR.

PR #229, opened 2016-06-21
Haproxy map support to speed up vhost to backend lookup.

ajays20078:
Currently vhost to backend lookup happens in the traditional way, where each rule is scanned to lookup the corresponding backend, this can add considerable latency when the number of apps/vhosts are large in number (especially for apps/vhosts which appear at the rear end of the rules comparison).

Ex:

frontend xyz
   {other_lines}
   use_backend backend1 if { hdr(Host) -i myapp.domain1.com }
   use_backend backend2 if { hdr(Host) -i myapp.domain2.com }
   ....
   ....
   ....
   use_backend backendN if { hdr(Host) -i myapp.domainN.com }

Here the lookups for domainN, N-1, N-2 etc . can take considerable time, since these comparisons are sequential. Hence this PR supports a `--haproxy-map` , which creates a lookup map which is supported by haproxy, so for the above example when we use haproxy-map, we get

frontend xyz
  {other_lines}
  use_backend %[req.hdr(host),lower,map_dom(/path/to/map)]

and /path/to/map contains:

myapp.domain1.com backend1
myapp.domain2.com backend2
...
...
...
myapp.domainN.com backendN,

which is a hash based lookup dict, this way domainN, N-1 , N-2 etc, lookups will be faster.

Currently it just works for host header and x-marathon-app-id header, it retains the rule based comparison for both path based routing and auth. So the config will be a hybrid of both when path based acls and auth are present.


brndnmtthws:
Woah, this is cool! Thanks for the PR. It'll take me some time to review it, so please hang tight.

marathon_lb.py
```diff
@@ -1268,6 +1397,9 @@ def get_arg_parser():
                         "number of tasks exposed via marathon-lb.",
                         type=int, default=1000
                         )
+    parser.add_argument("--haproxy-map",
+                        help="Use haproxy maps for domain name to backend"
```
brndnmtthws: Nit: prefer 'HAProxy' to 'haproxy'.

brndnmtthws:
Please add and document an API endpoint to fetch the map. Just copy the `:9090/_haproxy_getconfig` endpoint, and cal lit `_haproxy_getvhostmap` or something. See https://github.com/mesosphere/marathon-lb#api-endpoints.

marathon_lb.py
```diff
@@ -356,14 +359,27 @@ def config(apps, groups, bind_http_https, ssl_certs, templater):
...
-                hostname=app.hostname,
-                appId=app.appId,
-                backend=backend
-            )
+            if haproxy_map:
+                if 'map_http_frontend_appid_acl' not in duplicate_map:
+                    http_appid_frontend_acl = templater \
+                        .haproxy_map_http_frontend_appid_acl(app)
+                    http_appid_frontends += http_appid_frontend_acl.format(
+                        haproxy_dir="/".join(config_file.split("/")[:-1])
```
brndnmtthws: Precompute this directory and store it in a val. I'd also prefer if you used `os.path` instead of rolling your own.

marathon_lb.py
```diff
@@ -643,17 +661,31 @@ def generateHttpVhostAcl(templater, app, backend):
...
-            staging_http_frontends += http_frontend_acl.format(
-                cleanedUpHostname=acl_name,
-                hostname=vhost_hostname
-            )
+            if haproxy_map:
+                if 'map_http_frontend_acl' not in duplicate_map:
+                    http_frontend_acl = templater.\
+                        haproxy_map_http_frontend_acl_only(app)
+                    staging_http_frontends += http_frontend_acl.format(
+                        haproxy_dir="/".join(config_file.split("/")[:-1])
```
brndnmtthws: Ditto.

marathon_lb.py
```diff
@@ -682,14 +714,31 @@ def generateHttpVhostAcl(templater, app, backend):
...
-                        backend=backend
-                    )
+                    if haproxy_map:
+                        if 'map_https_frontend_acl' not in duplicate_map:
+                            https_frontend_acl = templater.\
+                                haproxy_map_https_frontend_acl(app)
+                            staging_https_frontends += https_frontend_acl.\
+                                format(
+                                    hostname=vhost_hostname,
+                                    haproxy_dir="/".
```
brndnmtthws: What's the hardcoded `"/"` mean?

marathon_lb.py
```diff
@@ -857,14 +907,28 @@ def generateHttpVhostAcl(templater, app, backend):
...
-                        hostname=app.hostname,
-                        appId=app.appId,
-                        backend=backend
-                    )
+                    if haproxy_map:
+                        if 'map_http_frontend_acl' not in duplicate_map:
+                            http_frontend_acl = \
+                                templater.haproxy_map_http_frontend_acl(app)
+                            staging_http_frontends += http_frontend_acl.format(
+                                haproxy_dir="/".
```
brndnmtthws: Same here.

marathon_lb.py
```diff
@@ -876,19 +940,35 @@ def generateHttpVhostAcl(templater, app, backend):
...
-                    appId=app.appId,
-                    backend=backend
-                )
+                if haproxy_map:
+                    if 'map_https_frontend_acl' not in duplicate_map:
+                        https_frontend_acl = templater.\
+                            haproxy_map_https_frontend_acl(app)
+                        staging_https_frontends += https_frontend_acl.format(
+                            hostname=app.hostname,
+                            haproxy_dir="/".join(config_file.split("/")[:-1])
```
brndnmtthws: Precompute.

marathon_lb.py
```diff
@@ -925,15 +1018,20 @@ def writeConfigAndValidate(config, config_file):
...
     else:
         logger.error("haproxy returned non-zero when checking config")
         return False
 
 
-def compareWriteAndReloadConfig(config, config_file):
+def compareWriteAndReloadConfig(config, config_file, map_array, haproxy_map):
     # See if the last config on disk matches this, and if so don't reload
     # haproxy
+    map_file = "/".join(config_file.split("/")[:-1]) + \
```
brndnmtthws: Precompute path.

marathon_lb.py
```diff
@@ -942,13 +1040,40 @@ def compareWriteAndReloadConfig(config, config_file):
...
-            logger.warning("skipping reload: config not valid")
+    if haproxy_map:
+        if not os.path.isfile(map_file):
+            open(map_file, 'a').close()
+        runningmap = str()
+        try:
+            logger.debug("reading map config from %s", map_file)
+            with open(map_file, "r") as f:
+                runningmap = f.read()
+        except IOError:
```
brndnmtthws: Hmm, so it continues if it fails to reload? What happens? Is there a possibility that HAProxy will be left in a weird state/

marathon_lb.py
```diff
@@ -1190,7 +1319,7 @@ def do_reset(self):
                                       self.__groups,
                                       self.__bind_http_https,
                                       self.__ssl_certs,
-                                      self.__templater)
+                                      self.__templater, self.__haproxy_map)
```
brndnmtthws: Last arg should be on a different line.

marathon_lb.py
```diff
@@ -1438,7 +1571,7 @@ def process_sse_events(marathon, config_file, groups,
                                    args.haproxy_config,
                                    args.group,
                                    not args.dont_bind_http_https,
-                                   args.ssl_certs)
+                                   args.ssl_certs, args.haproxy_map)
```
brndnmtthws: Last arg on new line.

marathon_lb.py
```diff
@@ -1453,4 +1586,4 @@ def process_sse_events(marathon, config_file, groups,
         # Generate base config
         regenerate_config(get_apps(marathon), args.haproxy_config, args.group,
                           not args.dont_bind_http_https,
-                          args.ssl_certs, ConfigTemplater())
+                          args.ssl_certs, ConfigTemplater(), args.haproxy_map)
```
brndnmtthws: One arg per line.

marathon_lb.py
ajays20078: Thats a line break for pep8 , 79 character limit, its actually

staging_https_frontends += https_frontend_acl.\
                                format(
                                    hostname=vhost_hostname,
                                    haproxy_dir="/".
                                    join(config_file.split("/")[:-1])
                                )

marathon_lb.py
ajays20078: Same as above.

marathon_lb.py
ajays20078: Reload logic is not changed, it still validates it in `writeConfigAndValidate` function.

marathon_lb.py
brndnmtthws: Oic.

marathon_lb.py
ajays20078: Went by the convention of other flags, also haproxy config flag is currently --**haproxy**-config, so used 'haproxy' instead of 'HAProxy'

brndnmtthws: I should have been more clear. I was only referring to the help message formatting.

marathon_lb.py
ajays20078: This is fixed in new commit.

ajays20078:
Changes added in new commit:

1. API endpoints to fetch maps, which gets enabled only when `haproxy-map` flag is set.
2. haproxy_dir variable is initialised once and being passed along so that it need not be re-computed in multiple places.
3. Config to load Lua module for vhostmap when `haproxy-map` flag is enabled.
4. Test cases and documentation to reflect all these.

config.py
```diff
@@ -92,6 +92,80 @@ def load(self):
 The head of the HAProxy config. This contains global settings
 and defaults.
 '''))
+        self.add_template(
```
brndnmtthws: Duplicating all of this violates the DRY principle, and it seems unnecessary. How about we just have the Lua endpoint enabled all the time? I think it's okay if it returns an empty response when the map isn't enabled.

README.md
```diff
@@ -150,6 +150,14 @@ You can skip the configuration file validation (via calling HAProxy service) pro
 $ ./marathon_lb.py --marathon http://localhost:8080 --group external --skip-validation
 ```
 
+### Using Haproxy maps for backend lookup.
+You can use haproxy maps to speed up web application (vhosts) to backend lookup. This is very useful for large installations where the traditional vhost to backend rules comparison takes considerable time since it sequentially compares each rule. Haproxy map creates a hash based lookup table so its fast compared to the other approach, this is supported in marathon-lb using `--haproxy-map` flag.
```
brndnmtthws: Change the first 'haproxy' to 'HAProxy'.

README.md
```diff
@@ -150,6 +150,14 @@ You can skip the configuration file validation (via calling HAProxy service) pro
...
+
+```console 
+$ ./marathon_lb.py --marathon http://localhost:8080 --group external --haproxy-map
+```
+Currently it creates a lookup dictionary only for host header(both http and https) and x-marathon-app-id header. But for path based routing and auth, it uses the usual backend rules comparison.
```
brndnmtthws: Space between 'header' and '(', use 'HTTP' and 'HTTPS' instead of lowercase versions, and prefer 'X-Marathon-App-Id'.

getvhostmap.lua
```diff
@@ -0,0 +1,48 @@
+-- A simple Lua script which serves up the HAProxy
+-- vhost to backend map file.
+function check_file_exists(name)
+   local f=io.open(name,"r")
+   if f~=nil then io.close(f) return true else return false end
+end
+
+function read_config_file(cmdline)
```
brndnmtthws: Can you replace instances of `config` with `vhostmap`? It's a little confusing when reading the code, because it implies that's `haproxy.cfg`.

brndnmtthws:
Okay, PR is looking good. I made a few more comments here and there.

config.py
ajays20078: Will returning HAProxy config itself instead of maps, when the flag is disabled work?

brndnmtthws: That seems confusing. I think an empty response is better.

ajays20078: The problem i see here with enabling the module all the time is:

-  Lets say initially the `haproxy-map` flag is disabled, you run marathon-lb and the lua module tries reading a non-existent file, so it returns empty response.
-  Now lets run marathon-lb with `haproxy-map` flag, so the map file get created and everything works as expected, the endpoint returns the map.
-  Run marathon-lb again without `haproxy-map` flag, the module looks for the file, which it finds(since it was created in step 2)  and it serves the map file even though the current config is not using the map.

ajays20078: The only workaround i can think right now to adhere to DRY as well as tackle is above use case is having the below line in base_config:

`lua-load /marathon-lb/getconfig.lua{LOAD_VHOST_MODULE}`
and then in marathon_lb

```
if haproxy_map:
    pass LOAD_VHOST_MODULE = "\n  lua-load /marathon-lb/getvhostmap.lua"
else:
   pass LOAD_VHOST_MODULE = ""
```


And similarly for endpoint Acl. Does that work?

brndnmtthws: How about we truncate the map every time MLB is started? This will address the scenario you pointed out, and it will simplify the logic.

brndnmtthws: I'm also not super concerned about the scenario you mentioned, because typically MLB will be run from inside a Docker container which only lives as long as the script is running. Any state inside the container is not preserved when it restarts.

ajays20078:
Changed the below things with new commit:

1. updated config and test_cases to load vhostmap lua module, irrespective of haproxy-map flag so that DRY priciple is not violated.
2. Update Lua module to handle the case where file is not created.
3. Fixed few formatting issues with readme.

brndnmtthws:
I did some testing, and it looks like it breaks if you specify vhosts with paths. The vhosts which have paths need to generate the host ACL. For example:

```
frontend marathon_http_in
  bind *:80
  mode http
  acl path_letsencrypt-dcos_10000 path_beg /.well-known/acme-challenge
  use_backend %[req.hdr(host),lower,map_dom(/marathon-lb/domain2backend.map)]
  use_backend letsencrypt-dcos_10000 if host_tweeter_mesosphere_com_letsencrypt-dcos path_letsencrypt-dcos_10000
```

You can see the acl for the `/.well-known/acme-challenge` path, but there's no ACL for `host_tweeter_mesosphere_com_letsencrypt-dcos`.

ajays20078:
Could you provide more details for debugging:

- What are the number of vhosts/apps in that example? Does any of those app have multiple vhosts?

-  Is it one path per vhost or multiple?

-  What is the corresponding config look like with `--haproxy-map` flag.

brndnmtthws:
Here are 2 app definitions I created to demonstrate this:

https://gist.github.com/brndnmtthws/0b3605d0925be73d8f1904aae95c0a59
https://gist.github.com/brndnmtthws/047a61ecbb482299294e5658c02ed260

And here's the resulting haproxy.cfg:

https://gist.github.com/brndnmtthws/07f9d694a0b4bc966ce040d16864a309

ajays20078:
This is fixed with new commit, it was the only case where it was not checking if the app doesn't contain path and auth, for rest of the cases it was implicit (control flows into `if haproxy_map` flag only when it doesn't contain path and auth in app for other cases, since `if app.path` and `if app.authRealm` conditions exists), but for multiple vhosts with path case, control used to flow into `if haproxy_map` block irrespective of presence of auth and path  in app. It is fixed in line 671 of [this](https://github.com/mesosphere/marathon-lb/pull/229/commits/4d23bd7c390524993a584c20bd11017b0e1e1218) commit.

Also the earlier test case in `test_config_haproxy_map_hybrid` had multiple vhost for one app and path based routing for another app. Have added another test case `test_config_haproxy_map_hybrid_with_vhost_path` where same app has both multiple vhosts and path based routing, so that this particular case is covered.

brndnmtthws:
Please run `./scripts/build-docs.sh` to regenerate the docs.

brndnmtthws:
It looks like there's another case where it fails due to a missing ACL for the HTTP to HTTPS redirect. I'll see if I can generate a config to reproduce it when I get a chance.

ajays20078:
- Regenerated the docs.

- It works fine for HTTP to HTTPS redirect acl, i have verified the code , have also added a test case for the same in `test_config_haproxy_map_hybrid_httptohttps_redirect`, so no issues here AFAIK.

brndnmtthws:
Excellent. Thank you for the PR and your patience.

PR #250, opened 2016-07-04
Blue green hybrid deployment with zdd

ajays20078:
Currently zdd supports either completely blue or completely green deployments, but there are usecases when both blue and green have to be live at the same time.

Ex: Lets say current deployment color is blue with 10 instances, there is a new deployment(green color) which first needs to be deployed on 2 instances (so traffic is split 80:20) and then if nothing breaks, green can be scaled up to full capacity.

This PR supports such kind of deployments with the help of new  flag `HAPROXY_DEPLOYMENT_NEW_INSTANCES` or `--new-instances` command-line-arg to zdd.py

How it works?

- When  `--new-instances` flag is specified, it creates only the specified number of instances in new app(and deletes the same number of instance from old app). So if old app has 10 instances and zdd is run with `--new-instances` as 2, we will end up with 8 instances of old app and 2 instances of new. This state is called hybrid state.
- When deployment group is in hybrid state, the only way to do new deployment is by converting it to complete current(`--complete-cur` flag) or complete previous(rollback using `--complete-prev`) first and then deploy new app(The new app can again have `--new-instances` flag).

So the lifecycle of deployment group will be:
                                                                  
Usual Single Color deployment ---`(--new-instances)`---> Hybrid deployment --`(--complete-cur or --complete-prev)`--> Single color.

Currently You can't use `--new-instances` flag when deployment group is in `hybrid` state.(Explanation for this is below)

Currently only one hop of hybrid is supported (ratio between blue:green is based on `--new-instances` flag initially set), and after that we can convert it to completely blue or completely green, but we cannot increase/decrease the hybrid ratio. [Ex: For the above case old:new deployment ratio is 80:20, after that we can scale it to either 100:0 or 0:100, but  it cant be scaled to 60:40 or 40:60 from 80:20). Currently updating `labels` in marathon triggers a new deployment so `HAPROXY_DEPLOYMENT_NEW_INSTANCES` cannot be updated dynamically since it will cause degradation as it doesn't wait for connections to drain. Once marathon patches [this](https://github.com/mesosphere/marathon/issues/3945) issue, we can support multiple hops in hybrid.

### DIFF
--- a/Longhelp.md
+++ b/Longhelp.md
@@ -1017,6 +1017,20 @@ The target number of app instances to seek during deployment. You
 generally do not need to modify this unless you implement your
 own deployment orchestrator.
                     
+## `HAPROXY_DEPLOYMENT_NEW_INSTANCES`
+  *per app*
+
+Specified as `HAPROXY_DEPLOYMENT_NEW_INSTANCES`.
+
+The number of instances to be created in new group while doing 
+blue-green deployments. You generally do not need to modify this
+unless you implement your own blue green deployment orchestrator.
+
+Ex: In zdd , if `haproxy_deployment_target_instances` is 10 and current
+deployment color is `blue`,  if you run zdd with `--new-instances` as 2,
+there will be 2 instances created in `green` group and 2 deleted from `blue`.
+So it will be blue:green with 80:20 ratio.
+
 
 ## `HAPROXY_{n}_GROUP`
   *per service port*
@@ -1149,6 +1163,3 @@ evaluated in the order it appears in the configuration.
 Ex: `HAPROXY_0_VHOST = 'marathon.mesosphere.com'`
 
 Ex: `HAPROXY_0_VHOST = 'marathon.mesosphere.com,marathon'`
-                    
-
-
--- a/README.md
+++ b/README.md
@@ -290,6 +290,35 @@ Zero downtime deployments are accomplished through the use of a Lua module, whic
 
 The ZDD script includes the ability to specify a pre-kill hook, which is executed before draining tasks are terminated. This allows you to run your own automated checks against the old and new app before the deploy continues.
 
+
+## Traffic splitting between Blue and Green Apps.
+
+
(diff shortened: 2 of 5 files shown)


brndnmtthws:
This is cool, thanks!

Can you a) add a note to README.md on this feature, and b) describe how you've tested it and that it doesn't affect existing behaviour. I will go over the code later today.

ajays20078:
a) Added note to README.md

b) Have added 3 new test cases (`test_hybrid`, `test_scale_new_app_instances_hybrid`, `test_complete_cur_exception`) which covers and tests new features along with old test cases. Also
Workflow enters new feature only when the new flags ( --new-instances, --complete-cur, --complete-prev) are set.

How to test it.

Use the workload `1-nginx.json` mentioned in tests directory.

**For normal workflow**
We can deploy app initially using the earlier command:

`zdd.py -j 1-nginx.json -m http://master.mesos:8080 -f -l http://marathon-lb.marathon.mesos:9090 --syslog-socket /dev/null`

and then change all the instances to new versions using the same command:

`zdd.py -j 1-nginx.json -m http://master.mesos:8080 -f -l http://marathon-lb.marathon.mesos:9090 --syslog-socket /dev/null`

These continue to work normally, since we don't specify the new flags.

**For new workflow**
if we need to test the new features, use the below commands (with new flags):

- To create a hybrid state with both apps use the following command:

`zdd.py -j 1-nginx.json -m http://master.mesos:8080 -f -l http://marathon-lb.marathon.mesos:9090 --syslog-socket /dev/null --new-instances 2`

- To convert hybrid state to completely current version instances use:
`zdd.py -j 1-nginx.json -m http://master.mesos:8080 -f -l http://marathon-lb.marathon.mesos:9090 --syslog-socket /dev/null --complete-cur`

We could use `--complete-prev` to convert it to completely previous version instances.

(We could also test it with `--resume` flag and with/without new flags)

brndnmtthws:
Superb, thank you.

PR #258, opened 2016-07-13
Added Custom Exception handling with exit status for various failure scenarios

ajays20078:
Added Exception handling to Zdd so that it throws different exceptions for different failure scenarios and exits with a different exit code for each of them. 

This makes life easier when zdd module or script is invoked automatically as different actions can be defined based on Exception Type/Exit Status.

### DIFF
--- a/zdd.py
+++ b/zdd.py
@@ -64,7 +64,7 @@ def marathon_get_request(args, path):
         response.raise_for_status()
     except requests.exceptions.RequestException:
         raise MarathonEndpointException(
-            "Error while quering marathon", url, traceback.format_exc())
+            "Error while querying marathon", url, traceback.format_exc())
     return response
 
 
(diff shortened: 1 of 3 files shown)


zdd.py
```diff
@@ -57,8 +59,12 @@ def query_yes_no(question, default="yes"):
...
 def marathon_get_request(args, path):
     url = args.marathon + path
-    response = requests.get(url, auth=get_marathon_auth_params(args))
-    response.raise_for_status()
+    try:
+        response = requests.get(url, auth=get_marathon_auth_params(args))
+        response.raise_for_status()
+    except requests.exceptions.RequestException:
+        raise MarathonEndpointException(
+            "Error while quering marathon", url, traceback.format_exc())
```
brndnmtthws: Should be 'querying'.

brndnmtthws:
This is great, thanks. One minor comment.

PR #261, opened 2016-07-13
Updated README for HAPROXY_0_VHOST usage.

vpack:
AWS ELBs use mixed case URLs and i had run into challenges when i cut and paste that into the json file. I had to troubleshoot with several haproxy configurations and identified this fix. I had also sent a comment to update the docs @ https://docs.mesosphere.com/1.7/usage/service-discovery/marathon-lb/usage/

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,8 @@ App labels are specified in the Marathon app definition. These can be used to ov
 }
 ```
 
+The virtual host name specified here must be in lower case.
+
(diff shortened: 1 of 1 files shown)


brndnmtthws:
Hmm, I appreciate the PR but the default hostname matching is case insensitive with MLB. Just grep for `hdr(host) -i` in https://github.com/mesosphere/marathon-lb/blob/master/Longhelp.md.

brndnmtthws:
I think I understand now why you had this problem. I created #278 to address it.

PR #270, opened 2016-07-19
use qdisc instead of iptables to prevent hanging

xiaods:
because iptables is block the sync packet, it caused 1s delayed.
use qdisc to prevent the hanging up.

### DIFF
--- a/service/haproxy/run
+++ b/service/haproxy/run
@@ -3,20 +3,6 @@ exec 2>&1
 export PIDFILE="/tmp/haproxy.pid"
 exec 200<$0
 
-addFirewallRules() {
-  IFS=',' read -ra ADDR <<< "$PORTS"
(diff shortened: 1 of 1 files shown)


brndnmtthws:
This is missing a few pieces. You need to set up the queues, and mark SYN packets so they actually get routed to the queues. In order to do it properly, the setup should be idempotent as well.

xiaods:
@brndnmtthws could u do me a favor, give some demo script?

xiaods:
@brndnmtthws thanks for your reminder, i will have a try with http://engineeringblog.yelp.com/2015/04/true-zero-downtime-haproxy-reloads.html script.

brndnmtthws:
Feel free to reopen this PR (or create another) if you feel it's ready for another review.

PR #273, opened 2016-07-21
Changing logic to always generate cert

anuruddhal:
As per the current code if we specify a  label as,
` HAPROXY_SSL_CERT=/etc/ssl/demo.pem `

Respective haproxy.cfg will be as following with two certs files.
```
frontend marathon_https_in
  bind *:443 ssl crt /etc/ssl/cert.pem crt /etc/ssl/wso2demo.pem
  mode http
```

But the /etc/ssl/cert.pem will not be generated and marathon-lb fails to start. This PR is to always generate certificate and fix the issue.

Else we can remove the logic that append the default cert. [1]

[1] https://github.com/mesosphere/marathon-lb/blob/master/run#L30

### DIFF
--- a/run
+++ b/run
@@ -45,8 +45,7 @@ if [ -n "${HAPROXY_SSL_CERT-}" ]; then
       SSL_CERTS+=",$certfile"
     fi
   done
-elif [ $ssl_certs_pos -eq 0 ]; then  # if --ssl-certs wasn't passed as arg to this script
-  # if no environment variable or command line argument is provided,
+
   # create self-signed ssl certificate
   openssl genrsa -out /tmp/server-key.pem 2048
   openssl req -new -key /tmp/server-key.pem -out /tmp/server-csr.pem -subj /CN=*/


brndnmtthws:
Hmm, can you describe exactly how you're running MLB? Are you setting an environment variable with `HAPROXY_SSL_CERT`? Are you passing the `--ssl-certs` flag?

anuruddhal:
I am passing both HAPROXY_SSL_CERT and --ssl-certs flag.

lloesche:
This PR breaks passing in a cert via the ENV as well as generation of a self signed cert when `HAPROXY_SSL_CERT` isn't set.

First it writes the ENV cert defined in `HAPROXY_SSL_CERT` to /etc/ssl/cert.pem in https://github.com/anuruddhal/marathon-lb/blob/aca08c3eb8e4a595985808cfee763986a98c5e46/run#L37

then it overwrites the same cert with a self signed one in https://github.com/anuruddhal/marathon-lb/blob/aca08c3eb8e4a595985808cfee763986a98c5e46/run#L53

Also note that https://github.com/anuruddhal/marathon-lb/blob/aca08c3eb8e4a595985808cfee763986a98c5e46/run#L50 will never be reached if https://github.com/anuruddhal/marathon-lb/blob/aca08c3eb8e4a595985808cfee763986a98c5e46/run#L35 doesn't evaluate true. I.e. if `HAPROXY_SSL_CERT` isn't set.

RE:
> I am passing both HAPROXY_SSL_CERT and --ssl-certs flag.

While using both the `HAPROXY_SSL_CERT` ENV var and `--ssl-certs` is supported I wonder why not use the one or the other? If you use `--ssl-certs` you have to ensure they are volume mounted into the container. At that point why also pass in certificates using the environment?

brndnmtthws:
Feel free to reopen this PR (or create another) if you feel it's ready for another review.

PR #283, opened 2016-07-30
Runsvdir

vixns:
Add runsvdir which run marathon_lb.py and haproxy.

This change allow to runit to auto restart marathon-lb.py when crashing or in poll mode, and to add extra services, in the docker image or by just mounting volumes under /marathon_lb/service/ directory.

I plan to add a prometheus haproxy_exporter as an optional service if/when this PR will be merged.


brndnmtthws:
test this please

PR #285, opened 2016-08-05
support DC/OS service accounts and HTTPS

alberts:


common.py
```diff
@@ -35,10 +43,44 @@ def set_marathon_auth_args(parser):
...
+        self.verify = False
+        if ca_cert:
+            self.verify = ca_cert
+
+    def __call__(self, auth_request):
+        payload = {
+            'uid': self.uid,
+            'exp': int(time.time()) + 3600,
+        }
+        token = jwt.encode(payload, self.private_key, 'RS256')
```
vespian: Shouldn't we get the encryption algorithm from JSON as well ?

Dockerfile
```diff
@@ -19,3 +16,9 @@ RUN apt-get update && apt-get install -y python3 python3-pip openssl libssl-dev
 COPY  . /marathon-lb
 
 WORKDIR /marathon-lb
+
+ENTRYPOINT [ "/marathon-lb/run" ]
```
vespian: Why did you change the order ?

PR #303, opened 2016-08-31
Use a separate HAProxy map file for mapping Marathon app IDs to backends

JayH5:
This has a few (small) advantages:
* The maps HAProxy loads are smaller
* The maps are easier to read/make sense of
* Slightly smaller/finer grained checks to see if config needs to be reloaded


JayH5:
I don't know if there's a tool that is used to regenerate the documentation?

ajays20078:
It is missing a API endpoint to fetch the app maps. (Lua module to fetch app maps, similar to [this](https://github.com/mesosphere/marathon-lb/blob/481520be6470fae2c6b46015a4b98782acab9800/getvhostmap.lua)) and also documentation for the API endpoint.
https://github.com/mesosphere/marathon-lb#api-endpoints

JayH5:
I noticed that, yes. It wouldn't be too hard for me to add something like a `:9090/_haproxy_getappmap` endpoint, but I wanted to be sure that that was something people needed or wanted?

brndnmtthws:
I think having the API endpoint is worthwhile. If someone plans to use this, I think there'd be an expectation that they'd be able to debug it somehow.

brndnmtthws:
test this please

JayH5:
Ready for review again. I've changed a few things:
* Added the `/_haproxy_getappmap` endpoint by rewriting `getvhostmap.lua` to `getmaps.lua`. (Sorry this is such a re-write.. took a while to wrap my head around the Lua).
* Reworked how HAProxy config is validated - before the config file was checked but with references to the old map files - now the temporary map files are written first and are set in the temp config before validation.
* Ran `build-docs.sh` to regenerate `LongHelp.md`

If there are integration tests for this stuff I'm not sure how to run them? I built & ran my changes manually and the config works as expected and the endpoints to fetch the maps work.

brndnmtthws:
test this please

PR #308, opened 2016-09-09
Python cleanup

JayH5:
General cleanup of Python code, mostly around imports and the test runner.

* Apply PEP-8 guidelines to imports: https://www.python.org/dev/peps/pep-0008/#imports
* Split development requirements into `requirements-dev.txt`
* Add note to README about how to run tests
* Use pytest to run tests
* Use pytest-cov for coverage tests
* Use pytest-pep8 for pep8 style check
* Fix some duplicated test names (this was causing at least 2 tests not to run)

This will obviously need someone to update the CI settings so that requirements are installed properly and all the tests are run.

There is one more problem with the code that I picked up running [`flake8`](http://flake8.readthedocs.io/en/latest/): [this method](https://github.com/mesosphere/marathon-lb/blob/v1.3.5/utils.py#L149-L150) doesn't actually set `ip_cache` as it sets `ip_cache` local to the method, not the module. I'd recommend running a linter like `flake8` as part of the build, but it's a separate command that would need to be added to the CI so I can't do that.


brndnmtthws:
test this please

PR #320, opened 2016-09-23
More Docker improvements

JayH5:
Minor tweaks plus..
* Install the runtime dependencies first so that that layer can be cached most of the time
* Be sure to install the latest Python packages even if a system package is already installed


brndnmtthws:
test this please

Issue #318, opened 2016-09-21
Ability to choose between '-sf' and '-st' HAProxy signal flags


h0tbird:
`marathon-lb` has the `-sf` flag hard-coded. Would you consider making it configurable?
I think this will be useful to people running long-lived TCP connections.

Ideally I would add a timeout to the `-sf` flag after which it will fail-over to the `-st` behavior. But this is a request for the upstream HAProxy project. What do you think?

brndnmtthws:
I'm not opposed to it, but I think I prefer #321.

h0tbird:
I think it will work for me too.

Issue #276, opened 2016-07-22
zombies


vixns:
I'm using latest docker image and  mesos containerizer.

On each topology change, the old haproxy process become a zombie.

```
19517 ?        Ss     0:00  |       |   \_ sh -c /marathon-lb/run sse --marathon ****
19519 ?        S      0:00  |       |   |   \_ /bin/bash /marathon-lb/run sse --marathon ****
19523 ?        S      0:00  |       |   |       \_ /usr/bin/runsv /marathon-lb/service/haproxy
19527 ?        S      0:00  |       |   |       |   \_ /bin/bash ./run
24062 ?        S      0:00  |       |   |       |       \_ sleep 0.5
19524 ?        Sl     0:00  |       |   |       \_ python3 /marathon-lb/marathon_lb.py --syslog-socket /dev/n
19590 ?        Zs     0:00  |       |   \_ [haproxy] <defunct>
23612 ?        Zs     0:00  |       |   \_ [haproxy] <defunct>
23658 ?        Ss     0:00  |       |   \_ haproxy -p /tmp/haproxy.pid -f /marathon-lb/haproxy.cfg -D -sf 662
```

brndnmtthws:
Oh dear. I wonder if d94b5fcab19c53922c9ce53f9964d31296dec71e or e36e8dbee3d0946cf3e8e17e0b7e3c8308390a15 introduced this.

brndnmtthws:
I checked all the MLBs in my soak cluster and I'm not seeing this:

```console
root@ip-10-0-6-34:/marathon-lb# ps waux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1  0.0  0.0  20256  3060 ?        Ss   15:46   0:00 /bin/bash /marathon-lb/run sse -m http://master.mesos:8080 --health-check --haproxy
root         8  0.0  0.0   4088   712 ?        S    15:46   0:00 /usr/bin/runsv /marathon-lb/service/haproxy
root         9  0.1  0.1 142676 23500 ?        Sl   15:46   0:00 python3 /marathon-lb/marathon_lb.py --syslog-socket /dev/null --haproxy-config /mar
root        10  0.0  0.0  20264  3068 ?        S    15:46   0:00 /bin/bash ./run
[...]

Issue #267, opened 2016-07-18
Race condition causing not released lock and too many HAProxy processes


mlowicki:
Hi,

Initially I've asked on Marathon [mailing list](https://groups.google.com/forum/?hl=en#!topic/marathon-framework/6wInFoQGu7w) but gather more data and since I've discovered that https://github.com/mesosphere/marathon-lb/blob/master/service/haproxy/run isn't provided by HAProxy but Marathon-lb then issue probably is caused by later.

Looks like lock isn't released and we've some hanging calls to flock:
```
root@lb2i:/marathon-lb/service/haproxy# ps aux | grep [f]lock
root       385  0.0  0.0   8464   868 ?        S    Jul13   0:00 flock 200
root       393  0.0  0.0   8464   856 ?        S    Jul13   0:00 flock 200
root       790  0.0  0.0   8464   856 ?        S    Jul13   0:00 flock 200
root      4951  0.0  0.0   8464   936 ?        S    Jul13   0:00 flock 200
root      6551  0.0  0.0   8464   856 ?        S    Jul13   0:00 flock 200
root     11841  0.0  0.0   8464   868 ?        S    Jul13   0:00 flock 200
root     11857  0.0  0.0   8464   864 ?        S    Jul13   0:00 flock 200
root     17515  0.0  0.0   8464   832 ?        S    Jul13   0:00 flock 200
root     17543  0.0  0.0   8464   796 ?        S    Jul13   0:00 flock 200
root     23167  0.0  0.0   8464   876 ?        S    Jul13   0:00 flock 200
root     24277  0.0  0.0   8464   856 ?        S    Jul13   0:00 flock 200
root     27799  0.0  0.0   8464   872 ?        S    Jul13   0:00 flock 200
root     27993  0.0  0.0   8464   880 ?        S    Jul13   0:00 flock 200
root     30333  0.0  0.0   8464   936 ?        S    Jul13   0:00 flock 200
root     34368  0.0  0.0   8464   884 ?        S    Jul13   0:00 flock 200
root     47233  0.0  0.0   8464   880 ?        S    Jul13   0:00 flock 200
root     48034  0.0  0.0   8464   936 ?        S    Jul13   0:00 flock 200
root     49056  0.0  0.0   8464   828 ?        S    Jul13   0:00 flock 200
```

```
root@lb2i:/marathon-lb/service/haproxy# flock --wait 5 ./run "echo d"
[...]

Issue #71, opened 2016-02-16
Stale HAProxy configurations remain listening


ejether:
We are seeing an interesting issue with v1.0.1 of marthon-lb running in SSE.
On a reconfiguration, occasionally the older process remains listening causing stale configuration and the  newer process to fail.

I have not be able to reliable reproduce it but it has cause several issues for us recently and the only way. It seems to occur when a service is flapping, or when many deployments are occurring at once. I have a theory that there is a race condition between when the new process has successfully started and the pidfile is read by the next starting process. (https://github.com/mesosphere/marathon-lb/blob/master/service/haproxy/run#L30)

I'm interested in any ideas anyone might have on how to reliably reproduce this or how to guaranty it doesn't bite us anymore. We are not ready to upgrade to v1.1.0 because we don't currently have an up front list of ports.

flosell:
Sounds a lot like issues we've had in the past. HAProxy, when doing a soft-reload, was spawning a new process with the new config and kept the old process around to finish processing requests currently in progress. But it was also accepting new requests, so the old and new config were both staying alive. Updating HAProxy to the most recent version fixed that issue for us. 

Along with that, at least on systemd, haproxy seems to have more issues reloading properly: 

* #5
* haproxy/haproxy#48

ejether:
We are using the marathon-lb docker container which is running HAProxy 1.5.8.
What version did you find fixed your problem @flosell?

flosell:
we are running HAProxy 1.5.14 right now

brndnmtthws:
When this occurs, can you check how many instances of HAProxy are running? As in, do a `ps` or `pidof haproxy`. Can you also check the output of `docker ps`? I think this may be similar to #72.

ejether:
I will do that @brndnmtthws:
I don't konw for sure if the problem is occurring on this node (its hard to find until something in particular starts throwing errors) but this is a typical output.
```
ejetherington@mesos-master-5vzd:~$ docker ps
CONTAINER ID        IMAGE                           COMMAND                  CREATED             STATUS              PORTS               NAMES
65c0b9df1fb6        mesosphere/marathon-lb:v1.0.1   "/marathon-lb/run sse"   7 days ago          Up 7 days                               marathon-lb
ejetherington@mesos-master-5vzd:~$ docker exec marathon-lb ps  aux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
[...]

PR #321, opened 2016-09-26
Optionally send SIGTERM after reload.

brndnmtthws:
This is to address issues #5, #71, #267, #276, and #318.


service/haproxy/run
```diff
@@ -38,7 +38,11 @@ reload() {
   socat /var/run/haproxy/socket - <<< "show servers state" > /var/state/haproxy/global
 
   # Trigger reload
-  haproxy -p $PIDFILE -f /marathon-lb/haproxy.cfg -D -sf $(pidof haproxy)
+  HAPROXY_PIDS=$(pidof haproxy)
+  haproxy -p $PIDFILE -f /marathon-lb/haproxy.cfg -D -sf $HAPROXY_PIDS
+  if [ -n "${HAPROXY_RELOAD_SIGTERM_DELAY-}" ]; then
+    sleep $HAPROXY_RELOAD_SIGTERM_DELAY && kill $HAPROXY_PIDS 2> /dev/null &
```
h0tbird: What is the purpose of `&` at the end of this line? The `sleep` runs in foreground and the `kill` runs in background, is that intended?

brndnmtthws: I don't believe that's the case. Test it yourself:

```
$ sleep 5s && echo finished &
```

lloesche: `&` and `&&` are both posix shell pipeline separators. `&&` executes sequentially until a nonzero status. `&` executes in parallel. If the purpose of the code is to wait `$HAPROXY_RELOAD_SIGTERM_DELAY` seconds before killing `$HAPROXY_PIDS` that's what it'll do, if it's parent `run` is still executing after the delay. If `run` was for whatever reason killed/restarted in the meantime the kill won't be executed.

h0tbird: So `sleep` and `kill` are sequentially executed but `kill` runs in parallel with the parent. I guess my real question is: How will this affect the overall time it takes to perform a `zdd` deploy?

brndnmtthws: That's right. If you set the delay to `5m`, then the old HAProxy processes will get reaped after 5 minutes, no matter what (and the deploy should proceed).

h0tbird: Ok, I was hoping for `zdd` to handle the killing of old HAProxy asynchronously. I don't know if this is doable because that might lead to a corrupted list of listeners.

ajays20078: - Is this PR applicable only when we explicitly specify the reload command? Because the reload function [here](https://github.com/mesosphere/marathon-lb/blob/f7ca0e03e6396b96fcd3f6b3464fb06ee3e33dc6/marathon_lb.py#L631) still uses default commands to reload if we don't specify anything explicitly.

- Also we need to explicitely specify the haproxy config file path as /marathon-lb/haproxy.cfg as per [this](https://github.com/mesosphere/marathon-lb/blob/f7ca0e03e6396b96fcd3f6b3464fb06ee3e33dc6/service/haproxy/run#L22) ?

brndnmtthws: Yeah, this PR will only affect people who run MLB using the Docker image, or the built-in run script. I'll fix the haproxy.cfg thing in another PR.

PR #329, opened 2016-09-30
Fix for IP per task in Marathon 1.3+.

### LINKED ISSUE #313: IP-Per-Container broken with Marathon-v1.3.0+
Latest Marathon changes make it [impossible to specify `discoveryInfo` and `ipAddress` fields](https://github.com/mesosphere/marathon/issues/4054). Instead, Marathon devs say to use `PortMappings`. 

Marathon-LB needs to update [where it expects this information](https://github.com/mesosphere/marathon-lb/blob/af45a72833de03ddb6db0f6fdb39506308ffef64/utils.py#L185) to be the new location.

Sort-of unrelated good news - if a user specifies a VHOST and `PortMappings.ContainerPort`, Marathon is now correctly allocating servicePorts in the PortMappings section. So [that issue in Marathon](https://github.com/mesosphere/marathon/issues/3636) is now effectively resolved.

I've been meaning to tackle these issues but now it seems less likely I'll get to them this week, so just wanted to start a log here of what's going on for the community.

brndnmtthws:
This should resolve #313.

[Test app 1](https://gist.github.com/brndnmtthws/677a715f2d2c7dec40c45dbd1a1296a6)
[Test app 2](https://gist.github.com/brndnmtthws/a9323c61e97564e2eb789eb88ae7b118)

[Results](https://gist.github.com/brndnmtthws/0cb5137ae0ec1922bc95100a7cb36566)

This can be tested by pulling the `fix-ip-per-task` Docker image tag.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -43,6 +43,65 @@ class TestUtils(unittest.TestCase):
 
         self.assertEquals(result, expected)
 
+    def test_get_task_ip_and_ports_ip_per_task_marathon13(self):
+        app = {
+            'ipAddress': {},
+            'container': {
+                'type': 'DOCKER',
+                'docker': {
+                    'network': 'USER',
+                    'portMappings': [
+                        {
+                            'containerPort': 80,
+                            'servicePort': 10000,
+                        },
+                        {
+                            'containerPort': 81,
+                            'servicePort': 10001,
+                        },
+                     ],
+                },
+            },
+        }
+        task = {
+            "id": "testtaskid",
+            "ipAddresses": [{"ipAddress": "1.2.3.4"}]
+        }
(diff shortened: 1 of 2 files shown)


djosborne:
Woo! Thanks for this @brndnmtthws ! I tried testing this yesterday with Calico but ran into troubles with my `app.json` (which turned out to just be typos on my end). Did a runthrough today with the v1.4.0 release and it works with calico networked docker tasks (even using VHOSTS), which is awesome!

Unfortunately, I found it does not work with CNI tasks, but I think the reason is that Marathon still is not auto-allocating servicePorts to CNI tasks. So will be raising a follow up report over with the Marathon team.

Thanks again!

PR #330, opened 2016-09-30
Memoize auth token.

brndnmtthws:


common.py
```diff
@@ -57,25 +57,32 @@ def __init__(self, credentials, ca_cert):
...
-        data = {
-            'uid': self.uid,
-            'token': token.decode('ascii'),
-        }
-        r = requests.post(self.login_endpoint, json=data, verify=self.verify)
-        r.raise_for_status()
-
-        auth_header = 'token=' + r.cookies['dcos-acs-auth-cookie']
-        auth_request.headers['Authorization'] = auth_header
+        if not self.auth_header or int(time.time()) >= self.expiry:
```
jgehrcke: About the condition `int(time.time()) >= self.expiry`: maybe add some tolerance interval here, so that we fetch the new authentication token maybe 10 minutes earlier than necessary, to prevent a type of face condition that could result in an unnecessary 401 response.

brndnmtthws: I don't think we need as much as 10 minutes, but yes good thinking. I added a 10s buffer.

jgehrcke: :+1: fine with me, and as we know clocks are never skewed! :)

common.py
```diff
@@ -57,25 +57,32 @@ def __init__(self, credentials, ca_cert):
...
-        r = requests.post(self.login_endpoint, json=data, verify=self.verify)
-        r.raise_for_status()
-
-        auth_header = 'token=' + r.cookies['dcos-acs-auth-cookie']
-        auth_request.headers['Authorization'] = auth_header
+        if not self.auth_header or int(time.time()) >= self.expiry - 10:
+            self.expiry = int(time.time()) + 3600
+            payload = {
+                'uid': self.uid,
+                'exp': self.expiry,
```
jgehrcke: @brndnmtthws I know this is confusing, but _this_ here is the lifetime of the service login token. It's fine if the service login token is valid just for a minute, because it only needs to survive the single login procedure.

In order to request an authentication token with a specific lifetime, you have to add an `exp` key to the `data` dict below!. Otherwise you get the default authentication token expiration time (5 days, currently).

brndnmtthws: Ah, I see. I merely rearranged the code. I'll update it accordingly.

brndnmtthws: See https://github.com/mesosphere/marathon-lb/pull/331

PR #331, opened 2016-09-30
Be more correct with auth token expiry.

brndnmtthws:
cc @jgehrcke


jgehrcke:
Haha, well, this looks good enough now :) So, LGTM! However, in an ideal world, I would now love to see a super-shortlived login token -- feel free to implement that :)

brndnmtthws:
Okay, updated accordingly.

jgehrcke:
:rocket:

PR #332, opened 2016-10-04
Correctly create haproxy chain

bakstad:
Don't use `pidof haproxy` as input to the `-sf` argument.

`pidof haproxy` can return multiple pids instead of the latest haproxy pid, which can break the reload chain. 

This will hopefully fix the problem where MLB hangs indefinitely and stops refreshing the haproxy config.

Added log message while waiting for new haproxy pid as we have seen the loop to run forever before. Without this print out the user knows nothing about the state of MLB.

Similar problems:
* https://groups.google.com/forum/#!topic/marathon-framework/R6RkBHaFtBo
* https://github.com/QubitProducts/bamboo/issues/79

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -660,6 +660,7 @@ def reloadConfig():
             subprocess.check_call(reloadCommand, close_fds=True)
             # Wait until the reload actually occurs and there's a new PID
             while len(get_haproxy_pids() - old_pids) < 1:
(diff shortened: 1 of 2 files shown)


brndnmtthws:
test this please

brndnmtthws:
Interesting. I changed it to use `pidof` because of how it was stated in the HAProxy documentation (http://cbonte.github.io/haproxy-dconv/1.6/management.html#3) which sounds to me like it's safe to pass multiple PIDs.

PR #333, opened 2016-10-05
Process events without second thread to get trid of deadlocks

bakstad:
An attempt to remove threading and complex locking logic when processing events.

Not necessarily a final solution, but want to discuss this way of doing things.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -1498,7 +1498,7 @@ class MarathonEventProcessor(object):
 
                 if not self.__pending_reset and not self.__pending_reload:
                     logger.info("19")
-                    if not self.__condition.wait(300):
+                    if not self.__condition.wait(30):
                         logger.info('condition wait expired')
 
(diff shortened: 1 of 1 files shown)


brndnmtthws:
I suspect your issue is probably resolved with 9ed32e3b5e48a5da1fc4ba21634d95a1b913457c.

bakstad:
I've been running 1.4.1 over night and it seems to solve the deadlock issue! 👍

I still don't see why we need an extra thread. Very easy to make mistakes on a important piece of software.

bakstad:
I can see that. Another approach could be to put the events on a queue (https://docs.python.org/3.5/library/queue.html) in the main thread, and in a separate thread have simple processing loop that reads from the queue. That would at least get rid of much of the locking logic.

That said, 1.4.1 is still going strong, so the deadlock seem to be gone.

bakstad:
Closing this as I don't have the time to look at it for a couple of weeks.

Issue #35, opened 2015-12-23
sse is not working anymore on Marathon v0.14.x


scalp42:
Using current master `4d521e1112af1aaaa73147fae886262107f1c6bf` and Marathon `0.14.0-1.0.442.ubuntu1404` (RC), I'm running into issues.

I modified the code to print events and you can see that nothing's happening:

```
 * Reloading haproxy haproxy                                                                                                                                                                                                                                             [ OK ]
marathon-lb: updating tasks finished, took 0.039763689041137695 seconds
marathon-lb: SSE Active, trying fetch events from from http://prod-marathon.gearzero.us:8080/v2/events
marathon-lb: {"remoteAddress":"10.92.32.192","eventType":"event_stream_attached","timestamp":"2015-12-24T00:06:46.432Z"}
marathon-lb: received event of type event_stream_attached
```

It looks like no events are coming in.

scalp42:
Here is another log.

The `date` command refers to when I scaled an app and I stopped the app because it was taking forever without any changes:

```
root@mesos-slave-i-309193f4 [gz-prod] /opt/marathon_lb # date
Wed Dec 23 23:59:12 UTC 2015
root@mesos-slave-i-309193f4 [gz-prod] /opt/marathon_lb # /opt/marathon_lb/bin/python3 marathon_lb.py --marathon http://prod-marathon.gearzero.us:8080 --haproxy-config /etc/haproxy/haproxy.cfg -s --group '*'
marathon-lb: overriding HAPROXY_HEAD from templates/HAPROXY_HEAD
marathon-lb: overriding HAPROXY_HTTP_FRONTEND_HEAD from templates/HAPROXY_HTTP_FRONTEND_HEAD
marathon-lb: setting default value for HAPROXY_HTTP_FRONTEND_APPID_HEAD
marathon-lb: overriding HAPROXY_HTTPS_FRONTEND_HEAD from templates/HAPROXY_HTTPS_FRONTEND_HEAD
marathon-lb: overriding HAPROXY_FRONTEND_HEAD from templates/HAPROXY_FRONTEND_HEAD
marathon-lb: setting default value for HAPROXY_BACKEND_REDIRECT_HTTP_TO_HTTPS
marathon-lb: setting default value for HAPROXY_BACKEND_HEAD
marathon-lb: setting default value for HAPROXY_HTTP_FRONTEND_ACL
[...]

PR #374, opened 2016-11-21
HAProxy 1.6.10

JayH5:

PR #381, opened 2016-12-14
Use PycURL for fetching event streams

toofishes:
Earlier issues #35 and #114 made things better in this department. However,
when running marathon-lb with a large (400+ applications) marathon instance,
there are still problems.

These problems can be traced back to Python itself, unfortunately:
http://stackoverflow.com/questions/21797753/efficiently-reading-lines-from-compressed-chunked-http-stream-as-they-arrive

Python requests uses urllib under the covers, and there are implicit issues
when 7.5 MB of JSON comes back on a single line, as we're seeing when certain
events are emitted. These events are deployment_info and deployment_success at
a minimum, there may be more.

By switching to PycURL, as noted in the Stack Overflow post, we bypass this
whole issue. We use an HTTP library that handles this particular edge case
well, reducing CPU usage dramatically when a large event comes in. It also
handles gzip compression, which means any 7.5 MB JSON dumps should shrink
significantly.

One unsolved problem remains here: the addition of DC/OS authentication support
in #285 is extremely tightly coupled to internal implementation details of the
python requests module. This simply won't work with this code, and I have zero
ability to fix or test it as we don't use DC/OS.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,10 @@ FROM debian:stretch
 RUN apt-get update && apt-get install -y --no-install-recommends \
         ca-certificates \
         iptables \
+        libcurl3 \
         liblua5.3-0 \
+        libssl1.0.2 \
+        openssl \
         procps \
         python3 \
         runit \
@@ -37,6 +40,7 @@ COPY requirements.txt /marathon-lb/
 RUN set -x \
     && buildDeps=' \
         gcc \
+        libcurl4-openssl-dev \
         libffi-dev \
         liblua5.3-dev \
         libpcre3-dev \
--- a/README.md
+++ b/README.md
@@ -86,14 +86,6 @@ newer versions.
 
 Syntax: `docker run mesosphere/marathon-lb sse [other args]`
 
-#### `event` mode
-**NOTE**: `event` mode is deprecated and will be removed from marathon-lb in future releases.
-
-In event mode, the script registers a HTTP callback in marathon to get
-notified when state changes.
-
-Syntax: `docker run mesosphere/marathon-lb event callback-addr:port [other args]`
-
 #### `poll` mode
 If you can't use the HTTP callbacks, the script can poll the APIs to get
 the schedulers state periodically.
--- a/common.py
+++ b/common.py
@@ -35,9 +35,8 @@ def setup_logging(logger, syslog_socket, log_format, log_level='DEBUG'):
 
 def set_marathon_auth_args(parser):
     parser.add_argument("--marathon-auth-credential-file",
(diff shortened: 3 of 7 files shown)


tylermarshall:
I have also been experiencing this issue. While the upgrades to HAProxy (https://github.com/mesosphere/marathon-lb/pull/374) have resolved some of the 100% CPU issues, I see giant spikes from Python itself from these calls.

brndnmtthws:
We're definitely going to need to support auth. It's probably worth dropping the other modes (polling & callback) as well.

It's probably okay to use PycURL for SSE, and requests for everything else. In which case, we can reuse the requests-based auth stuff, and just pass the token in the header when it's present.

toofishes:
@brndnmtthws awesome, thanks for the quick feedback! As long as the main thread of work here (switching to PycURL to work around issues in core python urllib) is acceptable to everyone, I can take some more time to get auth working again. I'll also remove the marked-as-deprecated old event subscription code to make this all a bit cleaner.

If I make a best-effort pass at restoring auth functionality, can I get someone with a DC/OS setup to test it? Your approach seems perfectly sane- continue to use requests to grab the token, but ensure the right token headers end up on the cURL-based request.

Here's the current situation, if it helps to show some real metrics on what this is trying to solve. The red background indicates deploy events in our environment; you can see these cause large CPU usage spikes by the marathon-lb docker process:

marathon-lb-cpu-load

brndnmtthws:
I'd be happy to test it and work with you on the PR. The auth code is here: https://github.com/mesosphere/marathon-lb/blob/master/common.py#L53-L90

It fetches a token periodically (approx. every hour), and sets the `Authorization` header with that token.

toofishes:
OK, @brndnmtthws - this PR has been updated to the best of my abilities. Changes since last time:

* DC/OS JWT-based auth should now be supported. However, I can't test it against an actual implementation, because I don't have one. Testing help would be welcome!
* The old event/listening mode has been dropped, and all code removed that was associated with it.
* I kept 'poll' around because it wasn't much of a burden.

Testing against a non DC/OS stack, this seems to be working fine for us. Our JSON blobs have grown to 8 MB at this point, and this implementation works much smoother than the native Python/requests one did.

toofishes:
Ping? Thoughts here?

If marathon 1.4.x is going to make this problem go away with smaller event payloads, this can likely get dropped, but we're still experiencing this problem all the time with the current version due to huge payloads.

brndnmtthws:
I apologize for taking so long to review this. I've got a lot of other stuff going on 🙂

I'll dig into it today, and test it w/ auth.

toofishes:
Thanks for taking a look and merging!

PR #383, opened 2016-12-16
add tini

vixns:


Dockerfile
```diff
@@ -8,11 +8,16 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         python3 \
         runit \
         socat \
-    && rm -rf /var/lib/apt/lists/*
+        gpg \
```
brndnmtthws: Shouldn't these be part of the build deps? I don't think they're required at runtime.

Dockerfile
```diff
@@ -29,22 +34,24 @@ RUN set -x \
         wget \
     ' \
     && apt-get update \
-        && apt-get install -y --no-install-recommends $buildDeps \
-        && rm -rf /var/lib/apt/lists/* \
+    && apt-get install -y --no-install-recommends $buildDeps \
```
brndnmtthws: 👍

Dockerfile
```diff
@@ -29,22 +34,24 @@ RUN set -x \
...
+    && rm -rf /var/lib/apt/lists/* \
 # Install Python packages with --upgrade so we get new packages even if a system
 # package is already installed. Combine with --force-reinstall to ensure we get
 # a local package even if the system package is up-to-date as the system package
 # will probably be uninstalled with the build dependencies.
     && pip3 install --no-cache --upgrade --force-reinstall -r /marathon-lb/requirements.txt \
     && /marathon-lb/build-haproxy.sh \
+    && gpg --keyserver ha.pool.sks-keyservers.net --recv-keys 595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7 \
```
brndnmtthws: 👍

Dockerfile
vixns: 👍
updated.

vixns:
tini and runit have different purposes. tini only prevent zombies in isolated environments, while runit is a (great) service supervisor.

brndnmtthws:
test this please

robsonpeixoto:
Why not use `docker run --init` ?

vixns:
Because we are using the mesos containerizer.

PR #385, opened 2016-12-16
add prometheus_exporter

vixns:


robsonpeixoto:
@vixns how to run the prometheus_exporter ? Or it will always run?

vixns:
always on.
I can add an environment var to control this.

brndnmtthws:
Is there a compelling reason for why this should be in the base MLB image? I think for something like this, it would be better to derive from the base MLB image and include your tooling. We could add a guide on how to do this, however.

vixns:
The prometheus exporter is highly coupled with haproxy, it make sense to include it in MLB.
If we don't, we have to expose the admin socket as tcp, which may not be a good idea in terms of security, that's why I propose to include it in MLB.

I can maintain an automated built derivative for this one, but it may be more convenient to integrate popular tools into the official MLB distribution, maybe by generating specific docker tagged images (slim, prom_exporter, ...).

Some available choices :
- Maintain several Dockerfiles in this repository that produce official tagged docker images
- Keep MLB as lean as possible, document and encourage community derivatives.

It's a political choice, up to you :)

brndnmtthws:
Let's leave this PR open for a bit and collect some feedback. If anyone feels strongly about it, this is the place to chime in.

JamieCressey:
For environments where Prometheus is not being used such functionality in the core image could be seen as bloat. I would be more in favour of extending the base Marathon-LB with such "add-ons".

vixns:
We should use separate projects for this kind of optional features.

https://github.com/vixns/mlb-kafka adds a prometheus exporter and rsyslog to send logs to kafka.
It's not FROM the official docker image for now, we'll switch back to it when https://github.com/mesosphere/marathon-lb/pull/460 will be merged.

mlb "flavored" projects may be listed in a specific section in the readme.

pgraemer:
I'am running this in a different marathon app just pointing it to marathon-lb. 

I'd love to see this in marathon-lb, but i'd say its not necessary to include it into marathon-lb.

PR #390, opened 2016-12-23
Emulate behaviour of `haproxy-systemd-wrapper`.

brndnmtthws:
For reference, see:
http://git.haproxy.org/?p=haproxy.git;a=blob;f=src/haproxy-systemd-wrapper.c;h=f6a9c852b277b5f193d6a952119a149b4cfe3e8b;hb=HEAD


vixns:
You may also care about signals propagation, see http://git.haproxy.org/?p=haproxy-1.7.git;a=commit;h=4351ea61fbddf88c960179d60b0e0f1b090f0b70

brndnmtthws:
Correct me if I'm wrong, but I think that only matters if you're using signals to reload HAProxy, which we aren't.

service/haproxy/run
```diff
@@ -33,7 +33,7 @@ reload() {
 
     # Trigger reload
     LATEST_HAPROXY_PID=$(cat $PIDFILE)
-    haproxy -p $PIDFILE -f /marathon-lb/haproxy.cfg -D -sf $LATEST_HAPROXY_PID 200>&-
+    /marathon-lb/haproxy_wrapper.py `which haproxy` -p $PIDFILE -f /marathon-lb/haproxy.cfg -D -sf $LATEST_HAPROXY_PID 200>&-
```
vixns: do not fork this, must block the execution until reload is over.
test the return code, should log to stderr in case of error.

service/haproxy/run
```diff
@@ -33,7 +33,7 @@ reload() {
...
     if [ -n "${HAPROXY_RELOAD_SIGTERM_DELAY-}" ]; then
       sleep $HAPROXY_RELOAD_SIGTERM_DELAY && kill $LATEST_HAPROXY_PID 200>&- 2>/dev/null &
     fi
```
vixns: HAPROXY_RELOAD_SIGTERM_DELAY  is no longer needed

haproxy_wrapper.py
```diff
@@ -0,0 +1,47 @@
...
+        pass
+
+
+def wait_on_haproxy_pipe(pipefd):
+    try:
+        ret = os.read(pipefd[0], 1)
+        if len(ret) == 0:
+            close_and_swallow(pipefd[0])
+            close_and_swallow(pipefd[1])
+            return False
```
vixns: is empty ret an error case ? if not should return True here

haproxy_wrapper.py
```diff
@@ -0,0 +1,47 @@
...
+if not pid:
+    os.environ["HAPROXY_WRAPPER_FD"] = str(pipefd[1])
+    # Close the read side
+    os.close(pipefd[0])
+    os.execv(sys.argv[1], sys.argv[1:])
+
+# Close the write side
+os.close(pipefd[1])
+while wait_on_haproxy_pipe(pipefd):
+    time.sleep(0.01)
```
vixns: add an explicit exit code / error message

brndnmtthws:
This seems to be working now. I'm going to go ahead and merge it soon.

PR #399, opened 2017-01-12
Enable auto-assign service port in zdd

mrkm4ntr:
I use marathon-lb with setting vhost and auto-assign service port.
But current zdd script force me to set a concrete servicePort and an alt port.


eodgooch:
I'm looking to pull this PR into my fork.. Based on what I am seeing you are assuming that the application without a service port is running in a docker container. That isn't always true.

Issue #402, opened 2017-01-16
Switch to an opt-in model for generating HAProxy configuration


vishnu2kmohan:
Say I have the following Marathon App definition where an app may listen on multiple ports but only requires one port `PORT0` to be exposed externally

```
{
  "id": "/pywebs",
  "cmd": "sh -c '(/opt/mesosphere/bin/python -m http.server $PORT0 &) && (/opt/mesosphere/bin/python -m http.server $PORT1 &) && (/opt/mesosphere/bin/python -m http.server $PORT2 &) && /opt/mesosphere/bin/python -m http.server $PORT3'",
  "cpus": 1,
  "mem": 128,
  "disk": 0,
  "instances": 1,
  "labels": {
    "HAPROXY_GROUP": "external",
    "HAPROXY_1_FRONTEND_HEAD": "",
    "HAPROXY_1_FRONTEND_BACKEND_GLUE": "",
    "HAPROXY_2_FRONTEND_HEAD": "",
    "HAPROXY_2_FRONTEND_BACKEND_GLUE": "",
    "HAPROXY_3_FRONTEND_HEAD": "",
    "HAPROXY_3_FRONTEND_BACKEND_GLUE": ""
  },
  "portDefinitions": [
    {
      "port": 0,
      "protocol": "tcp",
      "name": "web0",
      "labels": {
        "VIP_0": "/pywebs:8080"
      }
    },
    {
      "port": 0,
      "protocol": "tcp",
      "name": "web1",
      "labels": {
        "VIP_1": "/pywebs:8081"
      }
    },
    {
      "port": 0,
      "protocol": "tcp",
      "name": "web2",
      "labels": {
        "VIP_2": "/pywebs:8082"
      }
    },
    {
      "port": 0,
      "protocol": "tcp",
      "name": "web3",
      "labels": {
        "VIP_3": "/pywebs:8083"
      }
    }
  ],
  "healthChecks": [
    {
      "path": "/",
      "protocol": "HTTP",
      "portIndex": 0,
      "gracePeriodSeconds": 10,
      "intervalSeconds": 20,
      "timeoutSeconds": 10,
      "maxConsecutiveFailures": 3,
      "ignoreHttp1xx": false
    },
    {
      "path": "/",
      "protocol": "HTTP",
      "portIndex": 1,
      "gracePeriodSeconds": 10,
      "intervalSeconds": 20,
      "timeoutSeconds": 10,
      "maxConsecutiveFailures": 3,
[...]

Issue #403, opened 2017-01-16
Ensure that setting `--group "*"` matches only apps that have an HAPROXY_GROUP label defined and ignore apps that don't have it defined.


vishnu2kmohan:
Currently,  passing the command line argument `--group "*"` to Marathon-LB will cause it to advertise *all* apps on their service ports (provided there are no app and port-specific overrides present in their app definition), regardless of whether they have an `HAPROXY_GROUP` label present in their app definition.

This is not secure-by-default because there may be instances of apps that are *never* meant to be advertised because they explicitly did not set an `HAPROXY_GROUP` label.

Please alter the matching strategy to only advertise apps that have `HAPROXY_GROUP` set in their app definition.

This issue is related to #402 in that we'd like to enforce an opt-in model whenever possible.

brndnmtthws:
I've updated the docs in #410.

vishnu2kmohan:
Thanks!

PR #410, opened 2017-01-24
Introduce new `--strict-mode` option.

brndnmtthws:
This also adds a new `HAPROXY_{n}_ENABLED` option, which lets you
enabled/disable individual backends for an app.

This resolves issues #402 and #403.

### DIFF
--- a/Longhelp.md
+++ b/Longhelp.md
@@ -18,7 +18,8 @@ Marathon-lb just needs to know where to find Marathon.
 ```
 usage: marathon_lb.py [-h] [--longhelp] [--marathon MARATHON [MARATHON ...]]
                       [--haproxy-config HAPROXY_CONFIG] [--group GROUP]
-                      [--command COMMAND] [--sse] [--health-check]
+                      [--command COMMAND] [--strict-mode] [--sse]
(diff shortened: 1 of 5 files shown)


JohnOmernik:
I like this idea, but I think you've mentioned that for backwards compatibility, it's hard to switch enforcement.  The security guy in me wants to find a way to make this the default (strict mode) however, I understand the challenges of installed clusters... not sure how to best handle, perhaps with this change make some "marketing" hoopla? Update docs with notes recommending strict mode for new clusters? Recommending people make the change if they can, and then a page on understanding what needs to change?

brndnmtthws:
I'm not comfortable with enabling it by default at the moment. I think we can do a slow transition, but right now it would make a lot of people angry if their apps suddenly broke.

vishnu2kmohan:
LGTM

A blog post announcing `--strict-mode` with a plea to users to port their app definitions to a more secure version (along with a deprecation cycle - 1.5?) would be appreciated.

brndnmtthws:
I'm going to cut the next release as 1.5. I'll target enabling strict mode by default for 1.6.

PR #417, opened 2017-02-15
Python cleanup: make nosetests the default test runner, coverage, flake8

JayH5:
Fix issues reported by flake8, most importantly that the `set_ip_cache` function was not working.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -243,24 +243,33 @@ class CurlHttpEventStream(object):
 
 
 def resolve_ip(host):
-    cached_ip = ip_cache.get(host, None)
+    cached_ip = ip_cache.get().get(host, None)
     if cached_ip:
         return cached_ip
     else:
         try:
             logger.debug("trying to resolve ip address for host %s", host)
             ip = socket.gethostbyname(host)
-            ip_cache.set(host, ip)
+            ip_cache.get().set(host, ip)
             return ip
         except socket.gaierror:
             return None
 
 
-ip_cache = LRUCache()
+class LRUCacheSingleton(object):
+    def __init__(self):
+        self.lru_cache = None
+
+    def get(self):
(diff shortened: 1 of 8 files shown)


utils.py
```diff
@@ -243,22 +243,33 @@ def _split_lines_from_chunks(chunks):
...
+            ip_cache.get().set(host, ip)
             return ip
         except socket.gaierror:
             return None
-ip_cache = LRUCache()
 
 
-def set_ip_cache(new_ip_cache):
-    ip_cache = new_ip_cache
+class LRUCacheSingleton(object):
```
JayH5: This was really just the *easiest* solution I could think of. Definitely not the *best*.

JayH5: To be clear, the issue here is that in `set_ip_cache` the `ip_cache` variable is in the scope of the function, not the module. So `utils.ip_cache` was not actually being set, just a variable called `ip_cache` that was immediately thrown away.

brndnmtthws:
test this please

brndnmtthws:
Thank you very much @JayH5, this looks great.

🥇

robsonpeixoto:
Why `nosetests`?

JayH5:
@robsonpeixoto For me, pytest is my go-to test runner. A while ago I made a PR (#308) to switch to pytest in `requirements-dev.txt`/the README but it was never used for the Jenkins CI builds. I took this as a sign that the maintainers preferred nosetests :) The tests are pretty simple and don't use any functionality specific to a certain test runner/framework so whether it is nosetests or pytest doesn't make too much of a difference. So I thought I might as well make it consistent and use nosetests everywhere.

brndnmtthws:
I'll state for the record that I have no preference as to which test runner we use.

PR #430, opened 2017-03-14
use consistent line endings when comparing config

matt-deboer:
Since python replaces `\r\n` with `\n` when [reading files in text mode](https://docs.python.org/3.5/tutorial/inputoutput.html#reading-and-writing-files), any `\r\n` combinations provided in custom marathon labels will cause the configuration to always appear 'changed' when compared with the string read from the config file on disk.

This change assures consistent line endings are used during comparison to avoid unnecessary reloads.

Before this fix is applied, using this example config (from the wiki) or similar (containing `\r\n`) will cause the config to be seen as 'changed' every time a check is performed:
```
{
  "id":"app",
  "labels":{
    "HAPROXY_GROUP":"external",
    "HAPROXY_0_BACKEND_HEAD":"backend {backend}\r\n  balance {balance}\r\n  mode {mode}\r\n  timeout server 30m\r\n  timeout client 30m\r\n"
  }
}
```

### DIFF
--- a/config.py
+++ b/config.py
@@ -1165,7 +1165,7 @@ def set_balance(x, k, v):
 
 
 def set_label(x, k, v):
-    x.labels[k] = v
+    x.labels[k] = normalize_line_endings(v)
 
 
 def set_group(x, k, v):
@@ -1186,6 +1186,11 @@ def set_redirpath(x, k, v):
 
 def set_network_allowed(x, k, v):
     x.network_allowed = v
+
+# remove any line-ending inconsistencies
+# https://github.com/mesosphere/marathon-lb/pull/430
+def normalize_line_endings(v):
+    return v.replace("\r\n", "\n")
 
 
 class Label:
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -601,9 +601,7 @@ def config(apps, groups, bind_http_https, ssl_certs, templater,
     config += frontends
     config += backends
 
-    # python replaces '\r\n' with '\n' on read in text mode
-    # replace it here to get consistent comparison
-    return config.replace('\r\n', '\n')
+    return config
 
 
 def get_haproxy_pids():
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -601,7 +601,9 @@ def config(apps, groups, bind_http_https, ssl_certs, templater,
     config += frontends
     config += backends
 
-    return config
+    # python replaces '\r\n' with '\n' on read in text mode
+    # replace it here to get consistent comparison
+    return config.replace('\r\n', '\n')
 
 
 def get_haproxy_pids():


matt-deboer:
moved the config.replace into the config generation function.

IMO, fixing the docs (although important anyway) doesn't protect from falling into this trap. It may be repeated in some cases by copy-pasting config snippets from a windows box (which maybe is how they got into the docs in the first place?).

brndnmtthws:
@nlsun I've noted elsewhere that `\r\n` shouldn't be use, but that's not going to stop people from using it.

An alternative approach might be to call `split()` on the config in `compareWriteAndReloadConfig` and `compareMapFile`, which will result in the newlines being stripped (and makes which newline delimiters are used a moot point).

matt-deboer:
All, which change do you want to be made? The suggestion from @brndnmtthws sounds suspiciously like the original fix (which has been squashed out by now). I'm happy to do whatever you all think should be done, I'm just looking for a little consensus...

gpaul:
@matt-deboer The suggestion by @brndnmtthws (and the original fix) gets the job done, my suggestion aims for a bit more clarity and I believe "transform at the process boundary" is more python-esque. I would be happy with the compromise of @brndnmtthws' suggestion combined with a solid comment explaining why we are doing string substitution in the very strange place (and not, for example, replacing tabs) and a link to this PR.

brndnmtthws:
Just a minor correction on what I previously said: we should probably use `str.splitlines()`, rather than `str.split()`.

matt-deboer:
I was looking at the `1.9.0` release today, and I'm no longer able to repo.
Closing for now.

matt-deboer:
was a little premature in closing...it looks like maybe just the default check interval was changed so it took a little longer to show up. I'll update against current version.

matt-deboer:
ok; this has now been updated against 1.9, and removes inconsistencies at the process boundary (instead of replacing newlines in the resulting config file).
PTAL

justinrlee:
@gpaul @matt-deboer @drewkerrigan 
I know this is super old, but I'd like to clean up some of the marathon-lb PRs and issues.  I like the idea of this overall.  I'm happy to just merge as-is (either to `beta` or to `master`).  Thoughts?

gpaul:
@vespian wdyt? Like I said before, I'm happy to merge as is.

gpaul:
I had a chat to @vespian this morning and figured I'd summarize here.

Sorry, we dropped the ball on this PR.

The original maintainers of this project left Mesosphere a while ago.

Although ownership has passed to new members of the networking team at Mesosphere they are yet to grok the internals.

Given the weak status of CI and testing in this repository the new maintainers hesitate to merge patches that don't include extensive tests and were authored outside the team. Especially given that this doesn't appear to be a critical fix for an issue that is actively causing production outages. The perceived risk to merging is high.

In the short term the top priority with regards to this project is to introduce solid CI and run the existing tests, such as they are, on every PR.

Once we're running tests in CI on every PR we should add strong unit tests for this PR.
At that point I believe we will see it merged.

I hope you understand that the new maintainers do not feel comfortable merging a PR into a code base they're still unfamiliar with in the absence of good CI and testing.

Again, I'm sorry. We dropped the ball.

justinrlee:
Hey @matt.deboer.  So echoing what @gpaul said, we're working on getting some of the older PRs merged so we can move development forward a bit more.  In that vein, would it be possible to add some form of test case for this situation?  Thanks!

jkoelker:
ok to test

jkoelker:
@matt-deboer Again apologies for dropping the ball. I'll be merging this in, but currently it doesn't pass flake8. I've added commit https://github.com/mesosphere/marathon-lb/pull/622/commits/5f28d309a70f6ec5969475efec6681c45a0705e5 in https://github.com/mesosphere/marathon-lb/pull/622 that gets it to pass flake8 tests. If you'd like to incorporate that in this branch I'll merge it in, else I'll get another maintainer to merge in https://github.com/mesosphere/marathon-lb/pull/622 and close this one out.

jkoelker:
We've merged this in via https://github.com/mesosphere/marathon-lb/pull/622. Again apologies for the delay, and thanks for taking the time to make marathon-lb better!

PR #436, opened 2017-03-21
config: Fix missing labels referenced in overridable templates

nlsun:
A new template needs a corresponding add_template(), "setter", and if it
is "overridable", to append a label.

I also left a comment without modification when a template is missing
the "setter".


nlsun:
@lloesche @drewkerrigan please review.

PR #475, opened 2017-07-13
Add support for readiness checks

### LINKED ISSUE #467: Take readiness checks into consideration
Marathon-lb adds a service into the load balancer as soon as it sees the service is healthy. Marathon supports readiness checks as well, which can be used for running some tasks after deploy but before exposing the service. 

I think that Marathon-lb should also support Marathon's readiness checks, exposing the service via the load balancer only after the service is ready.

andresilva:
This PR adds support for [readiness checks](https://mesosphere.github.io/marathon/docs/readiness-checks.html). Similar to health checks, `marathon-lb` will wait until a task is ready before adding it as a backend.

Fixes #467.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -211,7 +211,7 @@ class Marathon(object):
     def list(self):
         logger.info('fetching apps')
         return self.api_req('GET', ['apps'],
-                            params={'embed': 'apps.tasks'})["apps"]
+                            params={'embed': ['apps.tasks', 'apps.readiness']})["apps"]
 
     def health_check(self):
         return self.__health_check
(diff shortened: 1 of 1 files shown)


marathon_lb.py
```diff
@@ -1407,6 +1407,24 @@ def get_apps(marathon):
...
 
+            if marathon.health_check() and 'readinessChecks' in app and \
+               len(app['readinessChecks']) > 0:
+                ready = False
+                if 'readinessCheckResults' not in app:
+                    # use previously cached result, if it exists
+                    if not readinessCheckResultCache.get(task['id'], False):
+                        continue
+                else:
+                    if len(app['readinessCheckResults']) == 0:
```
andrebeat: After a task is ready it seems that the `readinessCheckResults` array is always empty. Please confirm this is the correct behavior.

andresilva:
I'm closing this since it has been completely ignored and I have no intention of  spending any more effort on this project.

PR #476, opened 2017-07-14
Update for Marathon 1.5 API Changes

drewkerrigan:


tests/test_zdd.py
```diff
@@ -48,96 +62,104 @@ def test_scale_new_app_instances_up_50_percent(self, mock):
...
-        assert app['tasks'][2]['id'] not in results  # 2 l's d, 1 w/ scur/qcur
+        def run(apps):
+            listeners = _load_listeners()
+            haproxy_instance_count = 2
+            app = apps['apps'][0]
+            results = zdd.find_drained_task_ids(app,
+                                                listeners,
+                                                haproxy_instance_count)
+            from pprint import pprint
+            pprint(app['tasks'][0]['id'])
```
jdef: is this normally how test cases are debug-logged?

tests/test_zdd.py
```diff
@@ -329,11 +423,82 @@ def test_hybrid(self):
 
     @mock.patch('requests.get',
                 mock.Mock(side_effect=lambda k, auth:
+                          MyResponse('tests/zdd_app_blue_marathon1.5.json')))
+    def test_hybrid_marathon15(self):
```
jdef: what does "hybrid" refer to?

tests/zdd_app_blue_marathon1.5.json
```diff
@@ -0,0 +1,178 @@
+{
+  "apps": [
+    {
+      "id": "/nginx-green",
```
jdef: this file is called "blue" but the JSON is "green". intentional?

zdd.py
```diff
@@ -547,21 +565,15 @@ def set_app_ids(app, colour):
...
+    return set_service_port(app, servicePort)
 
 
 def select_next_port(app):
     alt_port = int(app['labels']['HAPROXY_DEPLOYMENT_ALT_PORT'])
-    if int(app['ports'][0]) == alt_port:
-        return int(app['labels']['HAPROXY_0_PORT'])
-    else:
-        return alt_port
+    if 'ports' in app:
```
jdef: marathon doesn't return the `ports` field anymore. use `portDefinitions` for `host` mode, or `container.portMappings` for `container` or `container/bridge` mode

tests/test_zdd.py
drewkerrigan: Nope, accidental

tests/test_zdd.py
drewkerrigan: So this is just the marathon15 version of `test_hybrid` which seems to be a slight modification of `test_simple`. The only difference I see is that `hybrid` increases the `HAPROXY_DEPLOYMENT_NEW_INSTANCES` to 1.

I have no idea where the `hybrid` name came from, but was attempting to refrain from completely refactoring.

tests/zdd_app_blue_marathon1.5.json
drewkerrigan: This is the marathon 1.5 version of `zdd_app_blue.json` which also contains an app called `/nginx-green` - I am assuming that the original author did this on purpose, although I'm not sure why.

zdd.py
drewkerrigan: As discussed, these changes are intended to be backwards compatible, so we'll still need to check the `ports` field in addition to the `portDefinitions`.

marathon_lb.py
```diff
@@ -1208,6 +1208,8 @@ def compareMapFile(map_file, map_string):
 
 
 def get_health_check(app, portIndex):
+    if 'healthChecks' not in app:
```
jdef: why is this change needed?

tests/test_utils.py
```diff
@@ -289,6 +358,38 @@ def test_ip_per_task_marathon13(self):
...
+                    'mode': 'container',
+                    'name': 'dcos'
+                }
+            ],
+            'tasks': [{
+                "id": "testtaskid",
+                "ipAddresses": [{"ipAddress": "1.2.3.4"}]
+            }],
+        }
+        self.assertEquals(self.assigner.get_service_ports(app),
```
jdef: ???

utils.py
```diff
@@ -135,7 +133,7 @@ def get_service_ports(self, app):
                                app.get('portDefinitions', []))
                            )
         ports = list(ports)  # wtf python?
-        if not ports and is_ip_per_task(app) and self.can_assign \
+        if not ports and mode == "container" and self.can_assign \
```
jdef: maybe a comment that this supports legacy IP/CT for Marathon 1.4.x and prior?

utils.py
```diff
@@ -275,25 +273,93 @@ def set(self, lru_cache):
...
+               .get('network', '')
+    if _mode == 'USER':
+        mode = 'container'
+    elif _mode == 'BRIDGE':
+        mode = 'container/bridge'
+
+    networks = app.get('networks', [])
+    for n in networks:
+        # Modes cannot be mixed, so assigning the last mode is fine
+        mode = n.get('mode', 'host')
```
jdef: I don't expect the default `host` to ever kick in here because Marathon would have normalized any items in `networks` to have a `mode`. That said, `host` isn't correct: if an empty network is specified (e.g. `networks: [ {} ]`) then the `mode` of that network is actually `container`

https://github.com/mesosphere/marathon/blob/596e86111e681e9923d7f751894ba62aaee64af5/docs/docs/rest-api/public/api/v2/types/network.raml#L73

utils.py
```diff
@@ -275,25 +273,93 @@ def set(self, lru_cache):
...
+
+    return mode
+
+
+def get_task_ip(task, mode):
```
jdef: in Marathon 1.5 a task may be assigned multiple containerIPs. the containerIP address you want may depend on which containerPort you're routing to

https://github.com/mesosphere/marathon/blob/596e86111e681e9923d7f751894ba62aaee64af5/docs/docs/rest-api/public/api/v2/types/appContainer.raml#L102

utils.py
```diff
@@ -275,25 +273,93 @@ def set(self, lru_cache):
...
+    if mode == 'host' or mode == 'container/bridge':
+        task_ports = task.get('ports')
+        if task_ports:
+            return task_ports
+        port_definitions = app.get('portDefinitions', [])
+        task_ports = [p['port']
+                      for p in port_definitions
+                      if 'port' in p]
+        if len(task_ports) == 0:
+            return None
```
jdef: for `container/bridge` mode the ports are in `portMappings` for Marathon 1.5, so doesn't this break? unless `get_app_task_ports` is only called for 1.4.x JSONs..

jdef: if `get_app_task_ports` is only called for marathon 1.4.x JSONs it might be nice to document that assumption

zdd.py
```diff
@@ -257,9 +257,13 @@ def find_drained_task_ids(app, listeners, haproxy_count):
     tasks = zip(get_svnames_from_tasks(app, app['tasks']), app['tasks'])
     drained_listeners = select_drained_listeners(listeners)
 
+    from pprint import pprint
+    pprint(drained_listeners)
+
     drained_task_ids = []
     for svname, task in tasks:
         task_listeners = [l for l in drained_listeners if l.svname == svname]
+        pprint(svname)
```
jdef: intentional debug?

zdd.py
```diff
@@ -257,9 +257,13 @@ def find_drained_task_ids(app, listeners, haproxy_count):
     tasks = zip(get_svnames_from_tasks(app, app['tasks']), app['tasks'])
     drained_listeners = select_drained_listeners(listeners)
 
+    from pprint import pprint
```
jdef: intentional debug?

zdd.py
```diff
@@ -496,24 +500,38 @@ def deploy_marathon_app(args, app):
...
-
+    container = app.get('container', {})
+    portMappings = container.get('docker', {}).get('portMappings', [])
+    if len(portMappings) > 0:
+        portMappings[0]['servicePort'] = int(servicePort)
+        app['container']['docker']['portMappings'] = portMappings
+        return app
+    portMappings = container.get('portMappings', [])
+    if len(portMappings) > 0:
+        app['container']['portMappings'][0]['servicePort'] = int(servicePort)
```
jdef: any reason why this isn't broken down into two statements like the prior `if` block?
```
portMappings[0]['servicePort'] = int(servicePort)
app['container']['portMappings'] = portMappings
```

just curious about the inconsistency and if there's a reason. in golang you get into trouble sometimes trying to assign things that live in a map value

zdd.py
```diff
@@ -496,24 +500,38 @@ def deploy_marathon_app(args, app):
...
+        return app
+    portDefinitions = app.get('portDefinitions', [])
+    if len(portDefinitions) > 0:
+        ['portDefinitions'][0]['port'] = int(servicePort)
```
jdef: missing `app` before `['portDefinitions...` here?

zdd.py
jdef: ok, so where's the check for `portDefinitions` or `portMappings`?

drewkerrigan: This function previously was not using either of those, so I assume that it could have worked before when `ports` was empty or unused.

marathon_lb.py
drewkerrigan: Because healthChecks may not be there which was causing a KeyError when this function was called without ensuring that 'healthChecks' was present.

tests/test_utils.py
jdef: nevermind

utils.py
drewkerrigan: I will change it to container.

utils.py
drewkerrigan: That is technically new functionality that I would prefer not to support with this PR. The primary goal here is to allow existing marathon apps to work with marathon-lb after upgrading to marathon 1.5. We can address feature enhancements separately.

utils.py
drewkerrigan: I think you missed `task_ports = task.get('ports')` - Note that for `container/bridge` it is checking the `task`'s ports, not `app`.ports.

zdd.py
drewkerrigan: No I only made the change to stay under 80 characters because of some lint checks. I resolved it another way in next commit.

utils.py
jdef: OK, maybe track this with a GH issue or JIRA then (and drop the link in this comment thread?)

utils.py
jdef: I see `app.get('portDefinitions', [])` in the above `if` block that executes in `container/bridge` mode. this will only work for 1.4.x

zdd.py
jdef: marathon 1.4.x and prior always returned `ports` for host or bridge mode, and also for docker containers in user mode. this is no longer the case: marathon 1.5.x never returns `ports` at all (because it's been deprecated forever and we finally stopped outputting it). this function is not compatible for and networking modes of marathon 1.5.x

utils.py
drewkerrigan: https://github.com/mesosphere/marathon-lb/issues/478

utils.py
```diff
@@ -275,25 +274,120 @@ def set(self, lru_cache):
...
-    """
-    Returns True if container network mode is set to USER
-    :param app:  The application to check.
-    :return:  True if using USER network, False otherwise.
-    """
-    c = app.get('container', {})
-    return c is not None and c.get('type', '') == 'DOCKER' and \
-        c.get('docker', {})\
-        .get('network', '') == 'USER'
+    _mode = app.get('container', {})\
```
jdef: TIL:
```
Python 2.7.8 (default, Apr 15 2015, 09:26:43) 
[GCC 4.9.2 20150212 (Red Hat 4.9.2-6)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> foo={'abc': None}
>>> foo.get('abc', {}).get('baz', '')
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'get'
```

so, when Marathon returns `null` for a value like `container` the logic here breaks down

jdef: (because the Python json parser translates a json `null` into `None`)

PR #480, opened 2017-07-28
marathon/json: properly handle app fields that may translate to None …

jdef:
…from a JSON null

* fixes https://jira.mesosphere.com/browse/DCOS-17521

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -126,11 +126,11 @@ class ServicePortAssigner(object):
                 if ports:
                     return list(ports)
 
-        ports = app.get('ports') or []
+        ports = app.get('ports', [])
         if 'portDefinitions' in app:
             ports = filter(lambda p: p is not None,
                            map(lambda p: p.get('port', None),
-                               app.get('portDefinitions') or [])
+                               app.get('portDefinitions', []))
                            )
         ports = list(ports)  # wtf python?
         # This supports legacy ip-per-container for Marathon 1.4.x and prior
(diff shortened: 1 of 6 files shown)


utils.py
```diff
@@ -126,11 +126,11 @@ def get_service_ports(self, app):
                 if ports:
                     return list(ports)
 
-        ports = app.get('ports', [])
```
nlsun: Instead of these, I would instead remove all None values from the initial JSON.

So something like:
```
Python 3.6.2 (default, Jul 17 2017, 16:44:45)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> json_value = '{"key":null,"key2":"yyy"}'
>>> def remove_nulls(d):
...      return {k: v for k, v in d.items() if v is not None}
...
>>> json.loads(json_value, object_hook=remove_nulls)
{'key2': 'yyy'}
```

And if we do this enough it may be worth it to have a wrapper for loading JSON

jdef:
I've successfully integration tested the changes made so far w/ oinker on DCOS 1.9.2 and 1.10-dev

utils.py
jdef: Your example doesn't look like it recursively removes None.

I'm about at the limit of my python skills here. If you want additional changes made here, I kindly request that you commandeer this PR @nlsun

nlsun: the `object_hook` should run on every unmarshalling of a JSON dict

```
Python 3.5.3 (default, Jul 26 2017, 15:11:40)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> json_value = '{"key":null,"key2":"yyy","key3":{"key4":null,"key5":"xxx"}}'
>>> def remove_nulls(d):
...     return {k: v for k, v in d.items() if v is not None}
...
>>> json.loads(json_value, object_hook=remove_nulls)
{'key3': {'key5': 'xxx'}, 'key2': 'yyy'}
```

asridharan:
Thanks @jdef @nlsun @drewkerrigan can we take over from here?

nlsun:
@mesosphere-ci retest this please

PR #481, opened 2017-07-31
exclude bad apps from generated config

matt-deboer:
This PR adds automatic sanitizing of configuration such that when an app containing invalid `HAPROXY_{n}_xxx` labels is introduced, it is excluded from configuration and a valid haproxy config is generated from the remaining apps.

The net result being that a single badly-configured application can no longer cause marathon-lb config updates to stop for all apps in the same cluster (or haproxy group).


matt-deboer:
Fixes #482

drewkerrigan:
@matt-deboer The changes look good, would you mind rebasing?

PR #492, opened 2017-09-05
Support providing MLB templates via environment variables rather than…

justinrlee:
… template tgz files.

https://jira.mesosphere.com/browse/DCOS-18367

### DIFF
--- a/config.py
+++ b/config.py
@@ -770,17 +770,25 @@ Use with HAPROXY_TCP_BACKEND_ACL_ALLOW_DENY.
         self.__load_templates()
 
     def __load_templates(self):
-        '''Loads template files if they exist, othwerwise it sets defaults'''
+        '''Look in environment variables for templates.  If not set in env, 
+        load template files if they exist. Othwerwise it sets defaults'''
 
         for template in self.t:
             name = self.t[template].full_name
-            try:
-                filename = os.path.join(self.__template_directory, name)
-                with open(filename) as f:
-                    logger.info('overriding %s from %s', name, filename)
-                    self.t[template].value = f.read()
-            except IOError:
-                logger.debug("setting default value for %s", name)
+            if os.environ.get(name):
+                logger.info('overriding %s from environment variable', name)
+                env_template_val = os.environ.get(name)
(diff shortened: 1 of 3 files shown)


nlsun:
It's rather difficult to squeeze text into environment variables because of the kind of escaping is required to get it to fit into JSON (e.g. `\\n`), so not terribly comfortable with this change. But at the same time it isn't exactly detrimental, I suppose.

justinrlee:
Yeah so I think I handled it so that it will accept either `\n` or `\\n` (or even `\\\n`, not sure on that one).  This is primarily to overwrite changes that only require a line or two.  But I've tested it as far as overwriting the HAPROXY_HEAD thing with no issues.

config.py
```diff
@@ -770,17 +770,25 @@ def __init__(self, directory='templates'):
...
-                filename = os.path.join(self.__template_directory, name)
-                with open(filename) as f:
-                    logger.info('overriding %s from %s', name, filename)
-                    self.t[template].value = f.read()
-            except IOError:
-                logger.debug("setting default value for %s", name)
+            if os.environ.get(name):
+                logger.info('overriding %s from environment variable', name)
+                env_template_val = os.environ.get(name)
+                # logger.info(env_template_val)
```
drewkerrigan: Looks like a leftover debug line? Please remove if so

config.py
```diff
@@ -770,17 +770,25 @@ def __init__(self, directory='templates'):
...
         for template in self.t:
             name = self.t[template].full_name
-            try:
-                filename = os.path.join(self.__template_directory, name)
-                with open(filename) as f:
-                    logger.info('overriding %s from %s', name, filename)
-                    self.t[template].value = f.read()
-            except IOError:
-                logger.debug("setting default value for %s", name)
+            if os.environ.get(name):
```
drewkerrigan: Would it be possible to add a test to `tests/test_marathon_lb.py`? You should be able to do something like

```
os.environ["key"] = "val"
```

justinrlee:
@drewkerrigan I've removed the debug line and added a basic unit test - can you validate?

Thanks!

PR #494, opened 2017-09-13
Fix return type of get_app_port_mappings

drewkerrigan:
This change addresses a bug introduced marathon-lb version 1.9.0 which affects apps using `USER` networking mode with containers as well as on DC/OS versions 1.9 and lower. Specifically apps with empty `portMappings`.

The encountered error looks similar to this:

```
File "/marathon-lb/utils.py", line 376, in get_port_mapping_ports
for p in port_mappings
TypeError: 'NoneType' object is not iterable
```

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -137,7 +137,7 @@ class ServicePortAssigner(object):
         if not ports and mode == "container" and self.can_assign \
                 and len(app['tasks']) > 0:
             task = app['tasks'][0]
-            _, task_ports = get_task_ip_and_ports(app, task)
+            task_ports = get_app_task_ports(app, task, mode)
             if task_ports is not None:
                 ports = [self._get_service_port(app, task_port)
                          for task_port in task_ports]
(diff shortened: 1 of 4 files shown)


utils.py
```diff
@@ -336,11 +336,11 @@ def get_app_port_mappings(app):
     portMappings = app.get('container', {})\
                       .get('docker', {})\
                       .get('portMappings')
-    if portMappings:
+    if portMappings is not None:
         return portMappings
 
     portMappings = app.get('container', {})\
-                      .get('portMappings')
+                      .get('portMappings', [])
```
nlsun: we should remove the default and then have a specific check to return `[]` if None. This is because `app` is may have None as the value.

We partially mitigated this elsewhere with the "cleanup_json" function, but there is one other place in the code that gets json (marathon.list()) which we don't sanitize

nlsun: we can either try to sanitize all entrypoints of json (like marathon.list()) or have this function also handle a `None` aka json null

drewkerrigan: Please check https://github.com/mesosphere/marathon-lb/pull/496 before we move forward with this PR

drewkerrigan:
@nlsun @jdef Ready for review.

drewkerrigan:
I've manually tested these fixes on DC/OS 1.9 and 1.10-beta2, merging and then releasing a new version.

PR #495, opened 2017-09-13
Remove None from util.py helpers

nlsun:


nlsun:
Code is too fragile, this is likely to cause more issues than it fixes.

PR #496, opened 2017-09-14
Fix ip cache usage, add cleanup_json

drewkerrigan:


drewkerrigan:
@nlsun I've fixed your branch - turns out that you accidentally broke the `resolve_ip` function, and I verified that all entry points into `utils.py` functions actually do go through `cleanup_json`, so I've added that to 2 of the tests with a comment explaining.

PR #502, opened 2017-09-22
Fix environment variables inconsistencies getting Marathon credentials from Vault

manriquecms:
There was some inconsistencies when using the Vault environment variables to get the Marathon credentials.

It seems there was a confuse between the required changes with the env VAULT_URL and the passed VAULT_HOST, VAULT_PORT, VAULT_PATH.

As you can see in the README.md file, now it's considered two environment variables:

VAULT_URL that is the vault service url (i.e: https://vault.service.mydomain:8200/v1), where is more flexible the declaration of the Vault URL (than the required HOST and PORT), maybe it couldn't has port because your DNS resolve it with a SRV entry, or change the api version, etc.

VAULT_MARATHON_PATH that is the root path where your user and password are located for marathon authentication, so if we need another path in the future for other purposes (i.e: certificates) we'll only define the concrete PATH in other environment variable.

Please consider this pull request and if you need me to change something, don't hesitate to request me.

Regards.


manriquecms:
I autoblock this PullRequest because maybe I'll include another fixes related

manriquecms:
Ok, now it's ready to the review, please if someone can do it, it would be great :+1:

mpenate:
Nice to have this fix! +1 for this

manriquecms:
I close this PR because my fork is in an unstable version

PR #505, opened 2017-10-05
Add container syslogd support and fix 'Waiting for Pids' race condition

drewkerrigan:
Corresponding config changes can be found here: https://github.com/mesosphere/universe/pull/1462

* Address various race conditions resulting in never-ending `Waiting for pid...`: Added these args:

```
  --retry-reload        Retry reload if unsuccessful after --reload-interval
                        seconds. (default: False)
  --reload-interval RELOAD_INTERVAL
                        When --retry-reload enabled, wait this long before
                        attempting another reload. (default: 10)
```
* Address `[ALERT] 278/182040 (164) : sendmsg logger #1 failed: No such file or directory (errno=2)` and `[ALERT] 278/182040 (164) : sendmsg logger #2 failed: No such file or directory (errno=2)`: Enable syslogd logging to container stdout (default to off) by setting the `HAPROXY_SYSLOGD` environment variable or `container-syslogd` value in `options.json` like so:
 
 ```json
   {
     "marathon-lb": {
       "container-syslogd": true
     }
   }
 ```

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -615,7 +615,8 @@ def get_haproxy_pids():
             "pidof haproxy",
             stderr=subprocess.STDOUT,
             shell=True).split()))
-    except subprocess.CalledProcessError:
+    except subprocess.CalledProcessError as ex:
+        logger.debug("Unable to get haproxy pids: %s", ex)
         return set()
 
 
@@ -649,8 +650,12 @@ def reloadConfig():
             old_pids = get_haproxy_pids()
             subprocess.check_call(reloadCommand, close_fds=True)
             # Wait until the reload actually occurs and there's a new PID
-            while len(get_haproxy_pids() - old_pids) < 1:
-                logger.debug("Waiting for new haproxy pid...")
+            while True:
+                new_pids = get_haproxy_pids()
+                logger.debug("Waiting for new haproxy pid (old pids: [%s], " +
+                             "new_pids: [%s])...", old_pids, new_pids)
+                if len(new_pids - old_pids) >= 1:
+                    break
                 time.sleep(0.1)
             logger.debug("reload finished, took %s seconds",
                          time.time() - start_time)
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -662,8 +662,9 @@ def reloadConfig():
                     break
                 timeSinceCheckpoint = time.time() - checkpoint_time
                 if (timeSinceCheckpoint >= reload_frequency):
-                    logger.debug("Still waiting for new haproxy pid after %s " +
-                                 "seconds (old pids: [%s], new_pids: [%s]).",
+                    logger.debug("Still waiting for new haproxy pid after " +
+                                 "%s seconds (old pids: [%s], " +
+                                 "new_pids: [%s]).",
                                  time.time() - start_time, old_pids, new_pids)
                     checkpoint_time = time.time()
                     if args.retry_reload:
@@ -1626,7 +1627,7 @@ def get_arg_parser():
                         help="Retry reload if unsuccessful after "
                         "--reload-interval seconds.", action="store_true")
     parser.add_argument("--reload-interval",
-                        help="When --retry-reload enabled, wait this"
+                        help="When --retry-reload enabled, wait this "
                         "long before attempting another reload.",
                         type=int, default=10)
     parser.add_argument("--strict-mode",
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -648,7 +648,8 @@ def reloadConfig():
         try:
             start_time = time.time()
             checkpoint_time = start_time
-            reload_frequency = 10 # Retry the reload every 10 seconds
+            # Retry or log the reload every 10 seconds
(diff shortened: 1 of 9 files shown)


drewkerrigan:
@urbanserj Was from another branch that I created this one from. I missed it when squashing other commits. I've removed it now.

marathon_lb.py
```diff
@@ -646,12 +647,32 @@ def reloadConfig():
...
             old_pids = get_haproxy_pids()
             subprocess.check_call(reloadCommand, close_fds=True)
+            new_pids = get_haproxy_pids()
+            logger.debug("Waiting for new haproxy pid (old pids: [%s], " +
+                         "new_pids: [%s])...", old_pids, new_pids)
             # Wait until the reload actually occurs and there's a new PID
-            while len(get_haproxy_pids() - old_pids) < 1:
-                logger.debug("Waiting for new haproxy pid...")
+            while True:
+                new_pids = get_haproxy_pids()
```
jdef: suggestion: move this to the end of the while loop to eliminate a couple of redundant calls

marathon_lb.py
```diff
@@ -646,12 +647,32 @@ def reloadConfig():
...
+                    logger.debug("Still waiting for new haproxy pid after " +
+                                 "%s seconds (old pids: [%s], " +
+                                 "new_pids: [%s]).",
+                                 time.time() - start_time, old_pids, new_pids)
+                    checkpoint_time = time.time()
+                    if args.retry_reload:
+                        logger.debug("Attempting reload again...")
+                        subprocess.check_call(reloadCommand, close_fds=True)
                 time.sleep(0.1)
+            new_pids = get_haproxy_pids()
```
jdef: suggestion: you can remove this as per the "move" suggestion above

marathon_lb.py
```diff
@@ -1602,6 +1623,13 @@ def get_arg_parser():
     parser.add_argument("--command", "-c",
                         help="If set, run this command to reload haproxy.",
                         default=None)
+    parser.add_argument("--retry-reload",
```
jdef: infinite retries? that smells like it could hide problems. maybe this could indicate a "max retries before failure"?

run
```diff
@@ -1,6 +1,27 @@
 #!/bin/bash
 set -euo pipefail
 
+log() {
+    msg="$1"
+
+    logline="[$(pwd) $0] $msg\n"
```
jdef: why is `$(pwd)` important for every log line?

run
```diff
@@ -1,6 +1,27 @@
...
+
+    printf "$logline" >&1
```
jdef: this seems pretty aggressive. why log to both stdout and stderr?

run
```diff
@@ -113,6 +134,12 @@ exec /marathon-lb/marathon_lb.py \
 EOF
 chmod 755 $LB_SERVICE/run
 
+log "Created $LB_SERVICE/run with contents:"
+LB_RUN=$(cat $LB_SERVICE/run)
+log ""
```
jdef: empty log lines? :(

service/haproxy/run
```diff
@@ -2,6 +2,15 @@
 exec 2>&1
 export PIDFILE="/tmp/haproxy.pid"
 
+log() {
+    msg="$1"
+
+    logline="[$(pwd) $0] $msg\n"
+
+    printf "$logline" >&1
```
jdef: ditto, why both stderr and stdout?

run
```diff
@@ -1,6 +1,27 @@
...
+
+if [ -n "${HAPROXY_SYSLOGD-}" ]; then
+    log "starting syslogd with container stdout"
+
+    # A hack to redirect syslog to stdout.
+    # This is necessary because haproxy only logs to syslog.
+    ln -sf /proc/$$/fd/1 /var/log/container_stdout
+
+    # Start syslog
+    /usr/sbin/syslogd -f "/marathon-lb/syslogd/syslog.conf"
```
jdef: .. and if it crashes (or is killed due to an OOM event) who restarts it? if you're going to run multiple daemons in a single container then you need a process manager (one w/ built-in support for dealing with zombies). anything else is probably not production-grade

marathon_lb.py
drewkerrigan: I'll change to `--max-reload-retries`

run
drewkerrigan: There are several files logging, and even multiple files named `run`, so the `$(pwd)` helps with debugging / context. I can pull it into a var like `LOG_PREFIX` so that it isn't run every time though.

run
drewkerrigan: Agree normal logs don't need to go to stdout, but it helps to have errors go to both stdout and stderr so that you get the context of what normal events happened before / after the error. I can make a separate `log_error` for that.

service/haproxy/run
drewkerrigan: I'll add a `log_error` here as well.

drewkerrigan:
@jdef PTAL - re: the syslogd comment - I've changed it so that syslogd is managed by runsvdir like the other daemons.

PR #506, opened 2017-10-13
Update base container to resolve CVE-2017-12424

keith-mcclellan:
Updated base container to debian:buster to resolve this attack vector.
https://security-tracker.debian.org/tracker/CVE-2017-12424


keith-mcclellan:
This is required by a gov't user - thanks!!

PR #510, opened 2017-10-28
Fix Build Status URI

orsenthil:


jkoelker:
Looks like this got fixed in https://github.com/mesosphere/marathon-lb/pull/548.

PR #515, opened 2017-11-19
send exceptions to sentry

ajays20078:
This PR adds sentry integration to mlb and allows us to send exceptions to sentry. This is enabled when using `--sentry` flag and sentry credentials needs to be configured in sentry.py

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -1561,10 +1561,12 @@ class MarathonEventProcessor(object):
         except requests.exceptions.ConnectionError as e:
             logger.error("Connection error({0}): {1}".format(
                 e.errno, e.strerror))
-            client.captureException()
+            if client:
+                client.captureException()
         except:
             logger.exception("Unexpected error!")
-            client.captureException()
+            if client:
+                client.captureException()
 
     def do_reload(self):
         try:
@@ -1721,7 +1723,8 @@ def process_sse_events(marathon, processor):
                 print(event.data)
                 print("Unexpected error:", sys.exc_info()[0])
                 traceback.print_stack()
-                client.captureException()
+                if client:
+                    client.captureException()
                 raise
     finally:
         processor.stop()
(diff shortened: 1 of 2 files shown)


sentry.py
```diff
@@ -0,0 +1,5 @@
+CONFIG = {
+    'sentry_user': "test_user",
+    'sentry_pwd': "test_pwd",
+    'sentry_endpoint': "http://test:8080"
```
jkoelker: These should all probably be environment variables.

marathon_lb.py
```diff
@@ -1729,6 +1738,19 @@ def load_json(data_str):
     # Process arguments
     arg_parser = get_arg_parser()
     args = arg_parser.parse_args()
+    if args.sentry:
+        import sentry
+        from raven import Client
+        global sentry_user
+        global sentry_pwd
+        global sentry_endpoint
+        global client
```
jkoelker: only `client` really needs to be a global, and it should be initialized to `None` somewhere.

marathon_lb.py
```diff
@@ -1729,6 +1738,19 @@ def load_json(data_str):
     # Process arguments
     arg_parser = get_arg_parser()
     args = arg_parser.parse_args()
+    if args.sentry:
+        import sentry
+        from raven import Client
```
jkoelker: This package needs to be installed in the requirements.txt file.

jkoelker:
Hi @ajays20078 Again apologies for the delay in someone reviewing this. I'm hesitant to merge this as its an integration with a specific vendor/provider. It would be much better if sentry would just scrape the docker logs and see the exceptions (although I've not looked into how sentry really works). That said the pr itself needs some work to function as is.

Since its been a long time, and I'm not sure if you're still interested in getting this merged, I'm going to close this. If you'd rather continue the discussion and update the PR, please don't hesitate to re-open this.

Issue #484, opened 2017-08-03
Caches are not cleaned up on Exception


ajays20078:
The finally block here, exists the processor but it doesn't cleanup the cache/lru, in case there is a exception ,leaving behind stale cache, is there a reason for not cleaning up the cache here?

https://github.com/mesosphere/marathon-lb/blob/0810a78ac317712a9e7275c4750fdda5f38052ae/marathon_lb.py#L1679

PR #516, opened 2017-11-19
exit on exception

ajays20078:
Exit mlb when ever it encounters a exception, so that all the caches and other variables are cleanup and it starts afresh. This fixes issue #484  as well.


jkoelker:
@ajays20078 Apologies for the delay in reviewing this PR.

I'd rather not `sys.exit()` anywhere outside of `__main__`. That said I believe that reseting the the healthCache on reconnect would be a more appropriate solution. However since it is a reconnect the cache should be updated with the current values. Please let me know if I'm missing something here and reopen this, but for now I think we should close this without merging.

PR #517, opened 2017-11-19
health_status_changed_event removed to reduce reloads

ajays20078:
Ignore `health_status_changed_event`, as it will be followed by `status_update_event` and mlb will eventually reload when status_update event is triggered. Also since HAProxy has its own health check, the task will be OOR on HAP so this event can be ignored. This reduces the number of reloads in a large setup and hence saves end user from frequent 5xx if there are containers dying frequently.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -1593,7 +1593,6 @@ class MarathonEventProcessor(object):
 
     def handle_event(self, event):
         if event['eventType'] == 'status_update_event' or \
-                event['eventType'] == 'health_status_changed_event' or \
                 event['eventType'] == 'api_post_event':
             self.reset_from_tasks()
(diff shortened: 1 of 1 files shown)


jkoelker:
@ajays20078 Apologies again for the delay in reviewing this PR.

Since this time we've moved to server side filtering of the events, and marathon itself has stopped sending the [`health_status_changed_event`](https://github.com/mesosphere/marathon/blob/2f87f35865d668f2b7b5abac678e13f112236512/changelog.md#health-reporting-via-the-event-stream). Although we still tell it to send us the `health_status_changed_event`, we can clean that up in another PR. Thank you for taking the time to make marathon-lb better, and again sorry for the long delay!

PR #518, opened 2017-11-19
print marathon master

ajays20078:
Log the current marathon host to which mlb is connected to when exception occurs. This helps in debugging issues in a large setup.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -161,6 +161,7 @@ class Marathon(object):
         self.__auth = auth
         self.__cycle_hosts = cycle(self.__hosts)
         self.__verify = False
+        self.current_host = None
         if ca_cert:
             self.__verify = ca_cert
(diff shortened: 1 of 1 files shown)


marathon_lb.py
```diff
@@ -226,7 +227,8 @@ def tasks(self):
         return self.api_req('GET', ['tasks'])["tasks"]
 
     def get_event_stream(self):
-        url = self.host + "/v2/events"
+        self.current_host = self.host
+        url = self.current_host + "/v2/events"
```
jkoelker: I'd rather we still use self.host here as there is another pending change that also touches this line and depending on ordering there will be conflicts.

jkoelker:
We've merged this in via #626. Again apologies for the delay and thank you for taking the time to make marathon-lb better!

PR #523, opened 2017-12-06
Fix reconnect logging and timing depending on error type

drewkerrigan:
Addresses the problem of continuously throwing the following error:

```
2017-11-03 05:31:36,031 marathon_lb: Caught exception
File "/marathon-lb/marathon_lb.py", line 1799, in <module>
process_sse_events(marathon, processor)
File "/marathon-lb/marathon_lb.py", line 1700, in process_sse_events
for event in events:
File "/marathon-lb/marathon_lb.py", line 239, in get_event_stream
for line in resp.iter_lines():
File "/marathon-lb/utils.py", line 241, in _split_lines_from_chunks
for chunk in chunks:
File "/marathon-lb/utils.py", line 226, in _iter_chunks
self._check_curl_errors()
File "/marathon-lb/utils.py", line 230, in _check_curl_errors
raise pycurl.error(*f[1:])
pycurl.error: (28, 'Operation too slow. Less than 1 bytes/sec transferred the last 300 seconds')
2017-11-03 05:31:36,043 marathon_lb: Reconnecting in 4.5s...
```

The fix is to reconnect immediately when this error is encountered. The changes also include a slight fix to the logging - it was not accurately displaying the number of seconds to wait previously.

After the fix, the log output for the above case is:

```
017-12-06 23:55:18,539 marathon_lb: Possible timeout detected: Operation too slow. Less than 1 bytes/sec transferred the last 300 seconds, reconnecting now...
```

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -41,6 +41,7 @@ from tempfile import mkstemp
 
 import dateutil.parser
 import requests
+import pycurl
 
 from common import (get_marathon_auth_params, set_logging_args,
                     set_marathon_auth_args, setup_logging, cleanup_json)
@@ -1160,8 +1161,9 @@ def truncateMapFileIfExists(map_file):
         os.ftruncate(fd, 0)
         os.close(fd)
 
+
 def generateAndValidateConfig(config, config_file, domain_map_array,
-                                app_map_array, haproxy_map):
+                              app_map_array, haproxy_map):
     domain_map_file = os.path.join(os.path.dirname(config_file),
                                    "domain2backend.map")
     app_map_file = os.path.join(os.path.dirname(config_file),
@@ -1169,7 +1171,6 @@ def generateAndValidateConfig(config, config_file, domain_map_array,
 
     domain_map_string = str()
     app_map_string = str()
-    runningConfig = str()
 
     if haproxy_map:
         domain_map_string = generateMapString(domain_map_array)
@@ -1178,15 +1179,13 @@ def generateAndValidateConfig(config, config_file, domain_map_array,
     return writeConfigAndValidate(
                 config, config_file, domain_map_string, domain_map_file,
                 app_map_string, app_map_file, haproxy_map)
-            
 
 
 def compareWriteAndReloadConfig(config, config_file, domain_map_array,
                                 app_map_array, haproxy_map):
-    
     changed = False
     config_valid = False
-    
+
     # See if the last config on disk matches this, and if so don't reload
     # haproxy
     domain_map_file = os.path.join(os.path.dirname(config_file),
@@ -1249,7 +1248,7 @@ def compareWriteAndReloadConfig(config, config_file, domain_map_array,
             changed = False
             config_valid = True
             logger.debug("skipping reload: config unchanged")
-    
+
     return changed, config_valid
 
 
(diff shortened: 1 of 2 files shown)


justinrlee:
Tested invalid haproxy config (based on the other #481 ), which looks like it behaves correctly.  Don't have a test case for the timeout issue, but the logic all looks correct to me.  If no feedback from @jdef or @GoelDeepak by Thursday evening, I'll merge (unless you wanna go ahead and merge now).

justinrlee:
Also, once this is merged, I'll cut a release (1.11.2)

drewkerrigan:
Also of note: I also fixed a bunch of flake8 problems introduced by the validation PR - not sure why they didn't fail for that. I think pyflakes got updated and now checks for additional things like "bare excepts."

Another fix was adding `build-essential` to the build deps because of this error: `x86_64-linux-gnu-gcc: error: /usr/share/dpkg/no-pie-compile.specs: No such file or directory` - That was preventing docker images from getting built.

marathon_lb.py
```diff
@@ -1891,21 +1891,44 @@ def load_json(data_str):
...
+                # 'Operation too slow. Less than 1 bytes/sec transferred
+                #  the last 300 seconds'
+                # In this case we should immediately reconnect
+                # without a backoff.
+                if errno == 28:
+                    m = 'Possible timeout detected: {}, reconnecting now...'
+                    logger.info(m.format(e_msg))
+                    currentWaitSeconds = 0
+                else:
+                    raise
```
jdef: does this `raise` trigger a reconnect? I'm not familiar with python exception handling

drewkerrigan: Great catch! This would actually exit the loop completely and not get caught by the following `except`. The intended behavior would be to log the exception and then reconnect with backoff as normal.

marathon_lb.py
```diff
@@ -1555,33 +1555,41 @@ def make_config_valid_and_regenerate(marathon, groups, bind_http_https, ssl_cert
...
+                     "took %s seconds",
+                     time.time() - start_time)
 
         return apps
 
     except requests.exceptions.ConnectionError as e:
         logger.error("Connection error({0}): {1}".format(
             e.errno, e.strerror))
-    except:
+    except Exception:
```
asridharan: Should we be just logging the exception as a string else won't be able to capture the reason. Same goes for other instances of this exception.

drewkerrigan: The line `logger.exception("Unexpected error!")` right after this one will actually print the exception as well as a stack trace if executed from within an exception context (as this is).

marathon_lb.py
```diff
@@ -1891,21 +1899,45 @@ def load_json(data_str):
...
             try:
                 process_sse_events(marathon, processor)
-            except:
+            except pycurl.error as e:
+                errno, e_msg = e.args
+                # Error number 28:
+                # 'Operation too slow. Less than 1 bytes/sec transferred
+                #  the last 300 seconds'
+                # In this case we should immediately reconnect
+                # without a backoff.
```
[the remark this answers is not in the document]
drewkerrigan: The error number 28 happens when there is no activity on the marathon event stream for 5 minutes. We're reconnecting immediately to ensure that the connection to marathon didn't die silently (see https://github.com/mesosphere/marathon-lb/blob/master/utils.py#L161-L170)

drewkerrigan:
@jdef @asridharan @justinrlee @GoelDeepak please check https://github.com/mesosphere/marathon-lb/pull/523/commits/9493c482c94380ee21eba1712334bf8e60fa01d5 - I've fixed a bug I noticed while working on this and testing it. The `processor.start()` was creating a new thread on every reconnect. Additionally the previous connection to marathon event stream was not getting closed.

marathon_lb.py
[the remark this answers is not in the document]
asridharan: Ah !! This makes a lot more sense now. Probably rewrite the comment to state what you just did in explicit terms. "Already waited for 5 minutes so we need to attempt a reconnection immediately".

marathon_lb.py
asridharan: Got it !!

marathon_lb.py
```diff
@@ -1891,21 +1883,66 @@ def load_json(data_str):
...
+                    else:
+                        logger.info("skipping empty message")
+            except pycurl.error as e:
+                errno, e_msg = e.args
+                # Error number 28:
+                # 'Operation too slow. Less than 1 bytes/sec transferred
+                #  the last 300 seconds'
+                # This happens when there is no activity on the marathon
+                # event stream for the last 5 minutes. In this case we
+                # should immediately in case the connection to marathon
```
asridharan: s/immediately/immediately reconnect

PR #527, opened 2017-12-09
Upgrade to haproxy 1.8.2

vixns:
* Use haproxy's new master worker, remove old iptables / sleep / lock hacks.

* :bangbang: BREAKING CHANGE: If you are using a custom HAPROXY_HEAD template, you _must_ remove "daemon" and add  "expose-fd listeners".

Require testing and a maybe way to auto fix custom HAPROXY_HEAD templates before merging.


justinrlee:
@vixns Aside from the lack of testing and HAPROXY_HEAD configurations, any reasons why you would consider this incomplete (WIP)?  Based on my rough testing, it looks like it works.  Thinking about cutting a beta build.

vixns:
haproxy 1.8 is still young and buggy with h2, we should wait for a few more minor versions. I'm still using it only on testing environments.
I think merging this PR may be sufficient as an alpha build, along a small release note.
More testing and documentation are required before reaching beta quality.

justinrlee:
FYI @vixns I'm merging this to a new branch.  Gonna add some quick handling of the `daemon` and `expose` stuff, and cut a build.

PR #528, opened 2017-12-11
enable strictmode

donghwicha:
Currently DCOS document says that strict mode is compatible with marathon-lb.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -104,7 +104,7 @@ class MarathonService(object):
         self.authUser = None
         self.authPasswd = None
         self.sticky = False
-        self.enabled = not strictMode
+        self.enabled = True
         self.redirectHttpToHttps = False
         self.useHsts = False
         self.sslCert = None


donghwicha:
it seems that strict mode is fully compatible with DCOS following document below, and it needs to be enabled. 

https://docs.mesosphere.com/1.10/networking/marathon-lb/usage/#mlb-strict-perm-install

justinrlee:
Hi Dongwhi, please verify whether you are using DC/OS enterprise edition or open source edition.

justinrlee:
@donghwicha can you clarify what you're trying to do with this?  Marathon-LB "strict" mode is distinct from DC/OS "strict" mode.  The former is a mechanism that changes the default behavior of Marathon-LB to not expose frontends unless they're explicitly exposed.  The latter is a security mode in DC/OS enterprise edition.  

If DC/OS strict mode is enabled, you're required to use a service account to authenticate Marathon-LB with DC/OS, but this has nothing to do with Marathon-LB strict mode.

donghwicha:
Oh I'm sorry for my late response. you can ignore this patch if you think it is not necessary.

PR #533, opened 2017-12-20
dont create frontend port binding for non tcp apps

ajays20078:
Dont create frontend port binding for non-tcp apps, as http and https works on headers.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -399,14 +399,26 @@ def config(apps, groups, bind_http_https, ssl_certs, templater,
             )
 
         frontend_head = templater.haproxy_frontend_head(app)
-        frontends += frontend_head.format(
-            bindAddr=app.bindAddr,
-            backend=backend,
-            servicePort=app.servicePort,
-            mode=app.mode,
-            sslCert=' ssl crt ' + app.sslCert if app.sslCert else '',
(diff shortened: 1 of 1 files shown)


jkoelker:
Hi @ajays20078, Sorry for the delay in getting around to reviewing this PR. I'm not sure I actually understand the use case. For now I'm going to close this, but if this is a feature you'd still like to see, please just reopen this PR, and drop a comment in here describing a bit more about why this is desired (preferably with example configs, but not required). Thanks for your work on making marathon-lb better.

ajays20078:
The use case is:

1. we don't need to use up a port for HTTP apps on MLB as http apps have vhost and clients use vhost to connect to mlb, instead of using host:port for HTTP. 

2. We dont reserve port ranges for mlb apps which can cause HAP reloads to fail when the same port is being used by HAP to proxy the request to upstream(as client port).

jkoelker:
Ah, I get it now. Thanks for clarifying. I'll test it out today, and add a commit updating the `Longhelp.md`.

> P.S. - I dont have permission to re-open a closed PR in this repo.

Sorry, I probably should have looked at the repo setting first ;).

jkoelker:
ok to test

PR #539, opened 2018-01-22
Use lightweight Marathon events.

jeschkies:
We introduced lightweight events with MARATHON_EE-1673. These do not
include the complete groups and make the events much smaller. This
should not only ease the consumption but also the workload for
Marathon.

This patch should close MARATHON-8036.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -226,7 +226,7 @@ class Marathon(object):
         return self.api_req('GET', ['tasks'])["tasks"]
 
     def get_event_stream(self):
-        url = self.host + "/v2/events"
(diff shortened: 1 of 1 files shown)


jeschkies:
Please note that this only works with the latest Marathon version.

drewkerrigan:
@jeschkies because this is such a widely used package, we need to maintain backwards compatibility. What is marathon's behavior when this query string is provided but not supported? (As would be the case in older marathon versions)

jeschkies:
@drewkerrigan, good point. I verified that it works with 1.4. The flag is just ignored.

jeschkies:
Dropped in favour of #541.

PR #541, opened 2018-01-25
Filter events on server side, in order to decrease load on Marathon

ichernetsky:
We introduced lightweight events with MARATHON_EE-1673. These do not
include the complete groups and make the events much smaller. This
should not only ease the consumption but also the workload for
Marathon. Furthermore Marathon LB should only request the events
it actually needs.

This patch should close MARATHON-8036 and MARATHON-8047.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -226,7 +226,7 @@ class Marathon(object):
         return self.api_req('GET', ['tasks'])["tasks"]
 
     def get_event_stream(self):
-        url = self.host + "/v2/events?" + \
+        url = self.host + "/v2/events?plan-format=light&" + \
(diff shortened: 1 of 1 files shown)


jeschkies:
Please checkout #539. We should also use the lightweight events.

jeschkies:
Dropped in favour of #541.

timcharper:
Should we remove the client side filtering eventType filtering logic now ? It seems redundant.

timcharper:
```
    def handle_event(self, event):
        # Server side filtering is set to event_type=api_post_event, event_type=health_status_changed_event and event_type=status_update_event"
        self.reset_from_tasks()
```

PR #548, opened 2018-02-07
Fix build status link

justinrlee:


justinrlee:
@drewkerrigan This is a relatively trivial doc change, but I think it adds some nice quality of life.  Mind adding an approval?

PR #556, opened 2018-02-23
Introduce "linear-increase" parameter to zdd.py.

DrMurx:
In our stack, due to some resource constraints, Marathon sometimes can't fulfill the scaling requests of `zdd.py`, especially since the final step (when number of new app instances has grown beyond number of old instances) might need a resource overcommitment by more than 150%.

To mitigate this, we introduced a `linear-increase` option to `zdd.py` which allows a continuous rollout, similar to a regular restart of the app in Marathon, but with the benefit of draining existing connections.

### DIFF
--- a/zdd.py
+++ b/zdd.py
@@ -413,11 +413,18 @@ def scale_new_app_instances(args, new_app, old_app):
        meet or surpass instances deployed for old_app.
        At which point go right to the new_app deployment target
     """
-    instances = (math.floor(new_app['instances'] +
-                 (new_app['instances'] + 1) / 2))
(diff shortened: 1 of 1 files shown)


justinrlee:
Hey @DrMurx sorry for the late response - gonna add this to my backlog, if that's okay.

@GoelDeepak @vespian FYI

GoelDeepak:
@justinrlee @vespian  Can we please capture this in jira so that it doesn't slide?

DrMurx:
Your CI bot is quite relentless 😅

Issue #554, opened 2018-02-13
disable app by app check


MikeMichel:
the app by app config check marks all our apps invalid because we have a custom HAPROXY_HTTPS_FRONTEND_HEAD to capture all let's encrypt challenges
```
frontend marathon_https_in
  bind *:443 ssl {sslCerts}
  mode http
  acl letsencrypt_challenge path_beg /.well-known/acme-challenge
  http-request allow if letsencrypt_challenge
  use_backend locksmith if letsencrypt_challenge
```
the backend locksmith does not exist when the app by app check is done so the config is invalid for every checked app.

the backend itself is a dynamic container so we can not hard code it via template.  a disable "app by app check" option for marathon-lb would help us here.

MikeMichel:
the patch from zliw is working. any chance to get it merged?

PR #558, opened 2018-03-06
disable configuration cleanup on failed validation (#554)

zliw:
This change set introduces an optional command line flag, which allows to skip the new app by app validation/cleanup in favour of the old behaviour:

A broken setup causes no reload of the haproxy.cfg. The old configuration is still usable. The haproxy.cfg is checked once entirely for configuration problems.

Unfortunately for the setup described in #554  the new app-by-app cleanup meant, that some applications would not be accessible anymore after removing them, depending on the order they are delivered by marathon. This currently prevents us from upgrading marathon-lb.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -1311,6 +1311,7 @@ def get_apps(marathon, apps=[]):
     # to a deployment group.
     processed_apps = []
     deployment_groups = {}
+
     for app in apps:
         deployment_group = None
         if 'HAPROXY_DEPLOYMENT_GROUP' in app['labels']:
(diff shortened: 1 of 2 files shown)


marathon_lb.py
```diff
@@ -1803,6 +1804,10 @@ def get_arg_parser():
     parser.add_argument("--skip-validation",
                         help="Skip haproxy config file validation",
                         action="store_true")
+    parser.add_argument("--skip-config-cleanup",
+                        help="If one app fails, don't try to make configuration "
```
jkoelker: Flake8: `./marathon_lb.py:1808:80: E501 line too long (81 > 79 characters)`

jkoelker:
@zliw Apologies for the delay in getting to this pr.

Would you mind rebasing this on master and changing the line to pass flake8? If so I'll merge this in.

jkoelker:
We've merged this in via https://github.com/mesosphere/marathon-lb/pull/625. Again apologies for the delay and thank you for taking the time to make marathon-lb better.

PR #562, opened 2018-03-23
Filter tasks by 'TASK_RUNNING' before adding to backend.

justinrlee:


marathon_lb.py
```diff
@@ -1475,6 +1475,11 @@ def get_apps(marathon, apps=[]):
                                task['id'])
                 continue
 
+            if task['state'] != 'TASK_RUNNING':
```
drewkerrigan: My only concern here is that there are several mesos task states that could happen other than `TASK_RUNNING` even if the task is actually still healthy: http://mesos.apache.org/documentation/latest/task-state-reasons/ / https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L2178-L2238

For example, if the mesos agent disconnects from mesos master, but all of the tasks on that agent are still healthy, the task might show up here as `TASK_LOST` because of `REASON_SLAVE_DISCONNECTED`. `TASK_UNKNOWN` can also happen for a variety of reasons even if the task is healthy.

We don't want to impact the uptime of a haproxy backend just because of some unrelated mesos connectivity problem.

A better idea might be to filter out task statuses that we know came from intentional removal of the task. Here are the ones I think we could safely remove:

* `TASK_KILLING`
* `TASK_KILLED`
* `TASK_FINISHED`
* `TASK_ERROR`

I'd be hesitant about ignoring `TASK_STAGING` / `TASK_STARTING` because if the task has a `host` and we find a port later in the check, then it is possible that the task will be available by the time the reload actually happens

jdef: Note: It's up to a framework whether or not to use TASK_STARTING at all, not much can be gleaned from it other than the maybe the task just isn't "fully operational" yet - whatever that means. So because it's optional, a task might live in TASK_STAGING for a while, even though it's able to service requests (because the framework hasn't yet returned TASK_RUNNING for some reason).
I guess if I'm a marathonLB admin and I know exactly which framework I'm fronting for, it might be useful to include/exclude these task states. Would it make sense to exclude these states based on a marathonLB flag?

justinrlee: Makes sense.  I'll modify behavior to exclude the 'safely removable ones'

justinrlee: I'm still iffy on including/excluding `TASK_STAGING`/`TASK_STARTING`, though.  Is it preferable to include or exclude tasks that haven't fully started?

Re: framework support, I think we only have to worry about Marathon.  My thought is that based on the documentation here: https://mesosphere.github.io/marathon/docs/task-handling.html#non-terminal-states, it's safe to exclude those two, but at this point maybe we'll wait for more reports of issues with non-fully-functional backends.

jdef: TASK_GONE is also a terminal state

marathon_lb.py
```diff
@@ -1475,6 +1475,12 @@ def get_apps(marathon, apps=[]):
                                task['id'])
                 continue
 
+            if task['state'] in ['TASK_KILLING', 'TASK_KILLED',
```
jdef: i'm no python expert, but does this basically end up iterating through an array to compare `task['state']` to each item in the "terminal state list"? if so, consider replacing the "terminal state list" with a `set(...)` fixture to get lookup times closer to O(1)

justinrlee:
This has been included with 1.12.0

PR #564, opened 2018-03-25
Work around python3 > dh-python > dkpg-dev > make dependency

justinrlee:
Somewhere upstream, a dependency was introduced such that:

`python3` depends on `dh-python`
`dh-python` depends on `dpkg-dev`
`dpkg-dev` depends on `make`

So uninstalling make (via `apt-get purge -y --auto-remove`) removes python3, and everything breaks.

This turns `make` from a build dependency to a runtime dependency - probably not ideal, but it's the easiest workaround for now.  We'll want to look at discussing how to change/prevent this in the future.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         python3 \
         runit \
         gnupg-agent \
-	socat \
+        socat \
+        make \
     && rm -rf /var/lib/apt/lists/*
 
 ENV TINI_VERSION=v0.13.2 \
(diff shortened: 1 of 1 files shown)


justinrlee:
@drewkerrigan See discussion (well, mostly me talking to myself here: https://mesosphere.slack.com/archives/C04GNUUFZ/p1521903316000057?thread_ts=1521840387.000363&cid=C04GNUUFZ)

Dockerfile
```diff
@@ -13,7 +13,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         python3 \
         runit \
         gnupg-agent \
-	socat \
+        socat \
```
justinrlee: `socat` change is just adding additional spaces to match the tabbing on other elements in this block.

PR #566, opened 2018-03-29
Append reload command only if it doesn't exist already

paambaati:
This PR includes a check for the reload command before appending it to the reload file. Without this, the `reload_haproxy.sh` file is slowly filled with multiple `sv reload ...` commands.

CC: @justinrlee @drewkerrigan @jdef


justinrlee:
Thanks, @paambaati!

PR #577, opened 2018-04-26
Update the documentation wrt. HAPROXY  connection mode

vespian:
Please check the JIRA issue for details: https://jira.mesosphere.com/browse/DOCS-2533

Issue #575, opened 2018-04-24
TCP proxy with HTTP healthcheck not working


fnerdwq:
I'm trying to setup a logstash service in  Marathon which is exposed via Marthon-LB. 
The proxy should TCP forward Port 514 to the container Port `syslog`, whereas the healthCheck should be a HTTP query on the container Port `api`.

The following example should - I think - logically do the job
```json
{
  "id": "/test/logstash",
  "cpus": 1.0,
  "mem": 1024,
  "instances": 1,
  "args": [
    "-e",
    "input { tcp { port => 10514 type => syslog } } output { stdout { codec => rubydebug } } }",
    "-w", "1",
    "-u", "5",
    "--log.level", "info"
  ],
  "container": {
    "type": "DOCKER",
    "docker": {
      "forcePullImage": false,
      "image": "docker.elastic.co/logstash/logstash-oss:6.2.3",
      "privileged": false,
      "network": "BRIDGE",
      "portMappings": [
        { "name": "syslog", "containerPort": 10514, "hostPort": 0, "protocol": "udp,tcp", "servicePort": 514 },
        { "name": "api", "containerPort": 9600, "hostPort": 0 }
      ]
    }
  },
  "healthChecks": [
    {
      "protocol": "MESOS_HTTP",
      "portIndex": 1,
      "path": "/",
      "ipProtocol": "IPv4",
      "gracePeriodSeconds": 300,
      "intervalSeconds": 60,
      "maxConsecutiveFailures": 3,
      "timeoutSeconds": 20
    }
  ],
  "labels": {
    "HAPROXY_0_GROUP": "external",
    "HAPROXY_0_MODE": "tcp",
    "HAPROXY_0_BACKEND_HEALTHCHECK_PORT_INDEX": "1"
  }
}
```
but the HAProxy backend config create is something like
```
backend test_logstash_514
  balance roundrobin
  mode http
  option forwardfor
  http-request set-header X-Forwarded-Port %[dst_port]
  http-request add-header X-Forwarded-Proto https if { ssl_fc }
  option  httpchk GET /
  timeout check 20s
  server 192_168_206_55_31504 192.168.206.55:31504 check inter 60s fall 4 port 31504
```
the Proxy Mode is changed to `http`!

Also the generated port check is wrong. Here port `31504` is checked, which maps to `syslog` . Instead port endpoint  `api` with `31505` should be used, since it's the Index 1 . See the tasks container configuration>
```
...
        "port_mappings": [
[...]

PR #581, opened 2018-05-02
Fix typos in the docs

vespian:
This is a follow-up for https://github.com/mesosphere/marathon-lb/pull/577


fnerdwq:
By chance I found this... Is this (and
#577) the "solution" to #575 ? 
Is there a chance that we can actually fix this issue instead of documenting it away? It would be great to have flexible health checking irrespectively of the actual proxy method.

PR #586, opened 2018-05-16
Make server ids persistent across reloads

vespian:
## High-level description

This PR introduces consistent IDs of backend servers. Thanks to this reloading HA-Proxy no longer results in backend server uptime data being reset - the backend server state is properly preserved across restarts in general and there is no downtime due to health checks needing to re-deploy.

There are two possible solutions to this approach:
* stateful
* hashing

The stateful approach requires a global state for each backend servers group which will be persisted across reloads. On top of that removal of backend servers need to be tracked and the corresponding server IDs need to be removed from the global state. 

The hashing approach is much simpler - there is a shared state but only during a reload, and the removals of backend servers are handled implicitly (i.e. given ID is simply not generated again). The disadvantage is that we can't use the full range of the server ID space - only ~30000 backend servers can be supported vs 32768 for the stateful approach. Still, I do not think this could be an issue (i.e. imagine haproxy handling a backend with 30000 backend servers).

Please check the docstrings introduced by the PR for more information and context for the introduced changes.

## Corresponding DC/OS tickets (obligatory)

 * https://jira.mesosphere.com/browse/DCOS-21644 `Marathon-LB 1.11 "re-inits" backend instances, making them appear offline for 30+ seconds. `

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -346,6 +346,93 @@ def mergeVhostTable(left, right):
     return result
 
 
+def calculate_server_id(server_name, taken_server_ids):
+    """Calculate a stable server id given server name
+
+        Calculates stable server id [1] for the given server name [2]
+        which has following properties:
+        * is unique/has not been assigned yet
+        * is an integer from the range 1-32767
+        * is stable - i.e. calling this function repeatably with the same
+          server name must yield the same server id.
+
+        THE STABILITY OF SERVER_ID IS GUARANTEED IF THE ORDER OF CALLS OF THIS
+        FUNCTION IS PRESERVED, I.E. THE BACKEND LIST IS SORTED BEFORE
+        PROCESSING
+
+        [1] http://cbonte.github.io/haproxy-dconv/1.8/configuration.html#5.2-id
+        [2] http://cbonte.github.io/haproxy-dconv/1.8/configuration.html#5.2
+
+        Args:
+            server_name(str): the name of the given backend server
+            taken_server_ids(set): list of allready assigned server ids
+
+        Returns:
+            An integer depicting the server ID
+    """
+    if server_name == '' or server_name is None:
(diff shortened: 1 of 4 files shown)


marathon_lb.py
```diff
@@ -346,6 +346,93 @@ def mergeVhostTable(left, right):
...
+        taken_server_ids.add(server_id)
+        return server_id
+
+    # We try to solve the collisions by recursively calling
+    # calculate_backend_id() with the server name argument set to the initial
+    # server name plus the calculated `server_name_shasum` appended to it.
+    # This way we should get stable IDs during the next haproxy
+    # reconfiguration.  The more backends there are the more likely the
+    # collisions will get. Initially the probability is 1/(2**15-1) * 100 =
+    # 0.03%. As the new_server_id gets longer the sha sum calculation will be
```
GoelDeepak: minor typo s/0.03%/0.003%

marathon_lb.py
```diff
@@ -550,6 +637,7 @@ def config(apps, groups, bind_http_https, ssl_certs, templater,
 
         do_backend_healthcheck_options_once = True
         key_func = attrgetter('host', 'port')
+        taken_server_ids = set()
```
GoelDeepak: would it make more sense to use dict()? or other similar data structure for fast lookup

vespian:
I have re-run the failed CI job variants for both `DC/OS EE Integration` and `
DC/OS Open Integration` tasks and they all succeded. The failures themselves were AWS CF stack formation failures. I think it is safe to assume that the failures were flakes/unrelated to the changes.

Unfortunately, it is required to have ALL the variants in the given CI run to have the green status checkmark on the PR - having everything green across two or more CI runs does not count :( I am leaning towards merging this PR as-is and addressing CI flakiness and refactoring at some point instead of re-running the CI until it gets green.

GoelDeepak:
Sounds fine to me

PR #595, opened 2018-08-01
#594 Follow redirects from non-leaders in Marathon 1.7

### LINKED ISSUE #594: Follow standby redirects
In Marathon 1.7.x, Marathon will be sending 302 redirect responses when requesting /v2/events from a non-leader. Marathon-lb needs to be updated to follow these redirects.

paambaati:
This PR fixes #594. We can land this once https://github.com/mesosphere/marathon/pull/6211 lands.

CC: @timcharper

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -38,7 +38,6 @@ from collections import defaultdict
 from operator import attrgetter
 from shutil import move, copy
 from tempfile import mkstemp
-from urllib import parse
 
 import dateutil.parser
 import requests
@@ -227,8 +226,11 @@ class Marathon(object):
         logger.info('fetching tasks')
         return self.api_req('GET', ['tasks'])["tasks"]
 
-    def get_event_stream(self, events):
-        url = self.host + "/v2/events"
+    def get_event_stream(self):
+        url = self.host + "/v2/events?plan-format=light&" + \
+              "event_type=status_update_event&" + \
+              "event_type=health_status_changed_event&" + \
+              "event_type=api_post_event"
         if events:
             url += "?" + urllib.parse.urlencode({'event_type': events}, doseq=True)
         return CurlHttpEventStream(url, self.__auth, self.__verify)
@@ -1795,10 +1797,6 @@ class MarathonEventProcessor(object):
         self.__pending_reload = False
         self.__haproxy_map = haproxy_map
 
-        self.relevant_events = ('api_post_event',
-                                'health_status_changed_event',
-                                'status_update_event')
-
         self.__thread = None
 
         # Fetch the base data
@@ -1900,7 +1898,9 @@ class MarathonEventProcessor(object):
         self.__condition.release()
 
     def handle_event(self, event):
-        if event['eventType'] in self.relevant_events:
+        if event['eventType'] == 'status_update_event' or \
+           event['eventType'] == 'health_status_changed_event' or \
+           event['eventType'] == 'api_post_event':
             self.reset_from_tasks()
 
     def handle_signal(self, sig, stack):
@@ -2081,7 +2081,7 @@ if __name__ == '__main__':
         while True:
             stream_started = time.time()
             currentWaitSeconds = random.random() * waitSeconds
-            stream = marathon.get_event_stream(processor.relevant_events)
+            stream = marathon.get_event_stream()
             try:
                 # processor start is now idempotent and will start at
                 # most one thread
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -38,6 +38,7 @@ from collections import defaultdict
 from operator import attrgetter
 from shutil import move, copy
 from tempfile import mkstemp
+from urllib import parse
 
 import dateutil.parser
 import requests
@@ -226,11 +227,10 @@ class Marathon(object):
         logger.info('fetching tasks')
         return self.api_req('GET', ['tasks'])["tasks"]
 
-    def get_event_stream(self):
-        url = self.host + "/v2/events?plan-format=light&" + \
-              "event_type=status_update_event&" + \
-              "event_type=health_status_changed_event&" + \
-              "event_type=api_post_event"
+    def get_event_stream(self, events):
+        url = self.host + "/v2/events"
+        if events:
+            url += "?" + urllib.parse.urlencode({'event_type': events}, doseq=True)
         return CurlHttpEventStream(url, self.__auth, self.__verify)
 
     def iter_events(self, stream):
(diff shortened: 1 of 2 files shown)


eLvErDe:
Hello,

Can you please re-open ?

It's not working properly: if authorization is enable it returns 401 with this patch, 302 otherwise

paambaati:
@eLvErDe A [401](https://httpstatuses.com/401)     just means that the credentials are invalid. Are you certain you’re using the correct credentials?

eLvErDe:
@paambaati 
No it's not a credential issue, it's pycurl not re-sending credentials on 302 code.

See: https://github.com/mesosphere/marathon-lb/issues/632 for a "suggestion"
But I don't understand this code at all, it's so complex for nothing... This should be re-written using an async framework, like aiohttp...

For the record, here's my implementation using aiohttp:

```python
"""
Listen to Marathon's SSE event bus, push message
to queue and process them with a queue worker
"""


# pylint: disable=line-too-long


import logging
import asyncio
import itertools
import random
import re
import json

import aiohttp

from consumer import SseEventConsumer


class SseEventListener(object):  # pylint: disable=too-many-instance-attributes
    """
    Listen to Marathon's SSE event bus, push message
    to queue and process them with a queue worker
    """


    def __init__(self, urls, registry_api_url, registry_api_key, username=None, password=None, timeout=15):  # pylint: disable=too-many-arguments

        assert isinstance(urls, list) and urls, 'urls must be list of non-empty string'
        assert isinstance(registry_api_url, str) and registry_api_url, 'registry_api_url must be a non-empty string'
        assert isinstance(registry_api_key, str) and registry_api_key, 'registry_api_key must be a non-empty string'
        for url in urls:
            assert isinstance(url, str) and url, 'urls must be list of non-empty string'
        if username is not None or password is not None:
            assert isinstance(username, str) and username, 'username and passord must be a non-empty strings o both None'
            assert isinstance(password, str) and password, 'password and passord must be a non-empty strings o both None'

        # Turn host list into a randomize never ending list
        # Access later with url property
        random.shuffle(urls)
        self.url_cycle = itertools.cycle(urls)
        self.active_url = None

        self.username = username
        self.password = password

        self.headers = {'Accept': 'text/event-stream'}
        self.auth = aiohttp.BasicAuth(self.username, self.password) if self.username else None

        self.logger = logging.getLogger(self.__class__.__name__)
        self.stopping = False
        self.timeout = timeout  # default value 15s, marathon seems to send empty line every 10

        self.sse_event = None
        self.sse_line_pattern = re.compile('(?P<name>[^:]*):?( ?(?P<value>.*))?')

        self.sse_event_queue = asyncio.Queue()
        self.sse_event_consumer_coro = None
        self.sse_event_consumer_task = SseEventConsumer(
            self.sse_event_queue,
            registry_api_url,
            registry_api_key,
        ).consume_forever()


    @property
    def url(self):
        """
        Property returning another url from a never ending shuffle list
        Also set self.active_url for logging purpose (ie)
        """

        self.active_url = self.url_cycle.__next__()
        return self.active_url


    def handle_sse_line(self, line):
        """
        Receive each SSE line and decide what to do
        """

        try:

            self.logger.debug('New message: %r', line)

            line = str(line, 'utf-8').strip()
            if not line:
                return

            sse_match = self.sse_line_pattern.match(line)

            if not sse_match:
                return

            name = sse_match.group('name')
            value = sse_match.group('value')

            if name == 'event':
                self.sse_event = value
            elif name == 'data':
                value = json.loads(sse_match.group('value'))
                self.logger.debug('SSE message %r with data %r', self.sse_event, value)
                if self.sse_event is not None:
                    self.sse_event_queue.put_nowait((self.sse_event, value))
                else:
                    self.logger.warning('Got a data message without an event one before, discarding')
            else:
                self.logger.warning('Unknown SSE message type %r', name)

        except Exception as exc:  # pylint: disable=broad-except
            self.logger.exception('SSE line handler crashed: %s: %s', exc.__class__.__name__, exc)


    async def listen_forever(self):
        """
        Listen SSE bus forever and switch server on error
        Read message and push them to self.
        """

        self.sse_event_consumer_coro = asyncio.ensure_future(self.sse_event_consumer_task)

        while not self.stopping:

            try:

                async with aiohttp.ClientSession(read_timeout=0) as session:
                    async with session.get(self.url, auth=self.auth, headers=self.headers) as response:
                        self.logger.info('Connecting to %s', self.active_url)
                        # Insane shit to override aiohttp.streams.DEFAULT_LIMIT
                        # It seems Marathon sends insanely long message with all apps definition sometimes
                        stream_reader = response.content
                        stream_reader._limit = 2 ** 20  # pylint: disable=protected-access
                        while not self.stopping:
                            # Replace async for so I can add a timeout on readline() here
                            # break patterns come from aiohttp code in streams.py
                            try:
                                line = await asyncio.wait_for(stream_reader.readline(), timeout=self.timeout)
                            except aiohttp.streams.EofStream:
                                self.logger.info('Received EofStream, server is closing')
                                break
                                # aiohttp issue I guess when closing
                            except AssertionError as exc:
                                if str(exc) == """yield from wasn't used with future""":
                                    self.logger.info('Aiohttp AssertionError catched  when closing, looks like a bug')
                                    break
                                raise
                            except asyncio.TimeoutError:
                                self.logger.error('Aiohttp readline() timed out after %d seconds', self.timeout)
                                break
                            if line == b'':
                                self.logger.info('Received empty line, server is closing')
                                break
                            self.handle_sse_line(line)

            except asyncio.CancelledError:
                pass

            except Exception as exc:  # pylint: disable=broad-except
                self.logger.exception('SSE bus listen crashed: %s: %s', exc.__class__.__name__, exc)

            finally:
                if not self.stopping:
                    self.sse_event = None
                    self.logger.info('Sleeping 1s before reconnecting to next host')
                    await asyncio.sleep(1)

        self.logger.info('Closing')


    def close(self):
        """ Stop processing """

        self.stopping = True
        self.sse_event_consumer_coro.cancel()
```

paambaati:
@eLvErDe Ah, thanks for the detailed info. A quick fix would be to include —

```
self.curl.setopt(pycurl.UNRESTRICTED_AUTH, True)
```

Can you test this? I’m not near a computer for a few days, so if you can confirm that this works, I can send a PR as soon as I’m available.

paambaati:
Oh, I see that @jkoelker already has https://github.com/mesosphere/marathon-lb/pull/633 with this fix!

@eLvErDe Can you test that branch out?

eLvErDe:
Yup, that seem to work correctly :-)

Issue #588, opened 2018-05-24
ZDD broken with the beta features merge #569


DrMurx:
#569 broke `zdd.py` - it's going to stuck in an endless loop:

Since haproxy isn't spawned as a daemon anymore, it always has at least 2 processes in the process list. Therefore, the check for the number of haproxy instances in method `check_haproxy_reloading` always returns a value greater than 1.

jkoelker:
fixed via https://github.com/mesosphere/marathon-lb/pull/598

PR #598, opened 2018-08-28
Increase ZDD pids from 1 to 2

Davidsoff:
referencing #588
since the container now runs 2 haproxy processes the check_haproxy_reloading function needs to be aware of this


jkoelker (approved):
LGTM

PR #600, opened 2018-09-19
updates build libcurl dependency from libcurl3 to libcurl4

GoelDeepak:


GoelDeepak:
@urbanserj  request for review

PR #606, opened 2018-12-05
Flake8 fixes

jkoelker:
Selectively uses raw strings to prevent escaping issues. Broken into two commits such that tests pass with each commit. Should fix the buildstatus.


jkoelker:
retest this please

jkoelker:
Closing in favor of #609.

PR #607, opened 2018-12-05
Fallback on next Marathon instance on exception

gpoulin:
If requests raise a TimeoutException it might be due to the current
instance being down. We should fallback to the next instance.

### Issue

When our first instance of marathon when down, marathon-lb never fallback on the other instances for `fetching apps`. Looking at the code it seems that marathon-lb doesn't cycle on marathon hosts if we requests raise an exception. This PR fix that by simply catching exception of requests and re-raising it if the request failed on all hosts.

Here is the log we got:
```
pycurl.error: (7, 'Failed to connect to {host1} port 8080: Connection refused')
2018-12-05 20:55:28,936 marathon_lb: Reconnecting in 4.5s...
2018-12-05 20:55:28,935 marathon_lb: fetching apps
2018-12-05 20:55:28,952 marathon_lb: Connection error(None): None
2018-12-05 20:55:28,953 marathon_lb: stopping event processor thread
2018-12-05 20:55:33,352 marathon_lb: starting event processor thread
2018-12-05 20:55:33,353 marathon_lb: SSE Active, trying fetch events from http://{host2}:8080/v2/events
2018-12-05 20:55:33,373 marathon_lb: received event of type event_stream_attached
2018-12-05 20:55:35,445 marathon_lb: received event of type event_stream_detached
2018-12-05 20:57:56,416 marathon_lb: received event of type status_update_event
2018-12-05 20:57:56,418 marathon_lb: received event of type instance_changed_event
2018-12-05 20:57:56,421 marathon_lb: fetching apps
2018-12-05 20:57:56,435 marathon_lb: Connection error(None): None
2018-12-05 20:57:56,465 marathon_lb: received event of type instance_changed_event
2018-12-05 20:57:57,562 marathon_lb: received event of type status_update_event
2018-12-05 20:57:57,563 marathon_lb: fetching apps
2018-12-05 20:57:57,565 marathon_lb: received event of type instance_changed_event
2018-12-05 20:57:57,566 marathon_lb: Connection error(None): None
```

### Disclaimer

I haven't try this code yet. I just wanted to know if there was something obvious I missed. If not we can try to reproduce the issue and see if that fix it.

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -166,27 +166,35 @@ class Marathon(object):
             self.__verify = ca_cert
 
     def api_req_raw(self, method, path, auth, body=None, **kwargs):
+        response, exception = None, None
         for host in self.__hosts:
             path_str = os.path.join(host, 'v2')
 
             for path_elem in path:
                 path_str = path_str + "/" + path_elem
+            try:
+                response = requests.request(
+                    method,
+                    path_str,
+                    auth=auth,
+                    headers={
+                        'Accept': 'application/json',
+                        'Content-Type': 'application/json'
+                    },
+                    timeout=(3.05, 46),
+                    **kwargs
+                )
 
-            response = requests.request(
-                method,
-                path_str,
-                auth=auth,
-                headers={
-                    'Accept': 'application/json',
-                    'Content-Type': 'application/json'
-                },
-                timeout=(3.05, 46),
-                **kwargs
-            )
+                logger.debug("%s %s", method, response.url)
+                if response.status_code == 200:
+                    break
+            except Exception as e:
+                logger.debug("%s %s", method, path_str, exc_info=True)
+                exception = e
 
-            logger.debug("%s %s", method, response.url)
-            if response.status_code == 200:
-                break
+        if not response:
+            assert exception is not None
+            raise exception
 
         response.raise_for_status()
 
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -167,7 +167,7 @@ class Marathon(object):
 
     def api_req_raw(self, method, path, auth, body=None, **kwargs):
         response, exception = None, None
-        for host in self.__hosts:
+        for i, host in enumerate(self.__hosts):
             path_str = os.path.join(host, 'v2')
 
             for path_elem in path:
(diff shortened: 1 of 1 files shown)


gpoulin:
We tested this and it fixed our issue.

marathon_lb.py
```diff
@@ -166,27 +166,35 @@ def __init__(self, hosts, health_check, strict_mode, auth, ca_cert=None):
...
-                    'Accept': 'application/json',
-                    'Content-Type': 'application/json'
-                },
-                timeout=(3.05, 46),
-                **kwargs
-            )
+                logger.debug("%s %s", method, response.url)
+                if response.status_code == 200:
+                    break
+            except Exception as e:
```
jkoelker: Can we catch a smaller set of exceptions? Perhaps just `requests.exceptions.Timeout`?

marathon_lb.py
```diff
@@ -166,27 +166,35 @@ def __init__(self, hosts, health_check, strict_mode, auth, ca_cert=None):
...
+                logger.debug("%s %s", method, path_str, exc_info=True)
+                exception = e
 
-            logger.debug("%s %s", method, response.url)
-            if response.status_code == 200:
-                break
+        if not response:
+            assert exception is not None
+            raise exception
```
jkoelker: I think refactoring this function in general would be better than storing the response and exception outside of the try block. Something like:

```python
class APIException(Exception):
    pass


def api_call(self, method, path, auth, body=None, **kwargs):
    for host in self.__hosts:
        try:
            response = request.request()
            if response.status_code != 200:
                continue
            response.raise_for_status()
            resp_json = cleanup_json(response.json())
            if 'message' in resp_json:
                response.reason = "%s (%s)" % (
                    response.reason,
                    resp_json['message'])

            return response

        except requests.exceptions.Timeout:
            continue

    raise APIException()
```

jkoelker: The idea is that it either returns a response or raises something. Since it's only catching the timeout, anything else will raise, but in the unlikely case that all hosts timeout, then the `APIException()` is raised.

marathon_lb.py
jkoelker: Pretty much everyone listed at https://docs.python.org/3/library/exceptions.html except `IOError`, as that's the basis for socket.error. Most everything else isn't specific to a host, e.g  `IndexError` or `AssertionError `. Those will be raised if there are issues in the code that gets run.

jkoelker:
Hi @gpoulin, I took the liberty to refactor it a bit and pushed it up as https://github.com/mesosphere/marathon-lb/pull/630. I want to do some manual testing tomorrow (still working on getting the CI all sorted out;), but wanted to let you know in the event you were interested in looking at it as well. If that refactor commit works for you and you would like to pull it into your branch, I will go ahead and merge this pr in once I've completed testing it as well.

jkoelker:
We've merged in #630. Thank you for taking the time to make marathon-lb better!

PR #582, opened 2018-05-04
#578 Remove temp map files only after checking for map config

paambaati:
This PR should fix #578 — I've included an additional check before removing temp map files.

CC: @cailian824, @drewkerrigan, @justinrlee

PR #615, opened 2019-02-16
Correctly validate global daemon flag

### LINKED ISSUE #613: Imprecise config checking prevents certain HAproxy configurations
[This line](https://github.com/mesosphere/marathon-lb/blob/master/marathon_lb.py#L459) is not a valid way to check if the ``daemon`` directive is being used.

I have a stanza I want to use in my HAproxy config:

```
  log /dev/log daemon notice
```

The ``daemon`` here is a syslog facility.

Because the config validation isn't very precise, it blocks me from starting MLB with such a configuration.

paambaati:
This should fix #613, reported by @jcmcken.

CC: @timcharper @brndnmtthws @drewkerrigan @justinrlee

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -456,7 +456,7 @@ def config(apps, groups, bind_http_https, ssl_certs, templater,
 
     # This should handle situations where customers have a custom HAPROXY_HEAD
     # that includes the 'daemon' flag or does not expose listener fds:
-    if 'daemon' in config or "expose-fd listeners" not in config:
+    if 'daemon' in config.split() or "expose-fd listeners" not in config:
         upgrade_warning = '''\
 Error in custom HAPROXY_HEAD template: \
 In Marathon-LB 1.12, the default HAPROXY_HEAD section changed, please \


timcharper:
@paambaati that does seem better. It's been a while since I've looked at this code; should we ideally arg-parse?

paambaati:
> should we ideally arg-parse?

@timcharper Yep.

timcharper:
I'll merge it right now; this seems like it at least makes it better. Thank you!

paambaati:
@timcharper While I have your attention, would you mind me asking a few questions?

1. Can you also merge #582? It is a simple enough fix.
2. What is the status of upgrades to HAProxy 1.9? I can confirm it works (we run it already on production on a medium-size cluster), but it would be nice to keep marathon-lb up to date.
3. There's been quite a few issues that aren't answered, and PRs that haven't been merged? What's the overall vision for the project looking like? Is there something contributors can help with?

mattj-io:
Hi @paambaati - hopefully you've seen #582 has also been merged, and we've bumped HAProxy to 1.8. Apologies that this repo hasn't had much love recently, but we're actively working to improve things and merge PR's that have been received - we would definitely love to have more external contributors !

PR #626, opened 2019-03-22
 Rebase and Flake8 fixes of #518

jkoelker:


jkoelker:
Just a trivial rebase as both this and #624 touched `get_event_stream`, no matter which one was merged first the other conflicted.

Issue #559, opened 2018-03-08
"Using marathon-lb" link broken in README.md


lishim:
Link points to Mesosphere website, which returns 404.

PR #627, opened 2019-03-25
readme: point to github wiki

jkoelker:
Links `Using marathon-lb` to a page on the github wiki instead of
`docs.mesosphere.com` so the links can be updated without a PR when the
URLs change.

Fixes: #559

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ fast, efficient, battle-tested, highly available load balancer with many advance
 
 ### Getting Started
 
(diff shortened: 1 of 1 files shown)


urbanserj:
There are some useful docs [here](https://docs.mesosphere.com/services/marathon-lb/).

jkoelker:
> There are some useful docs [here](https://docs.mesosphere.com/services/marathon-lb/).

Thanks I'll link that on the wiki page as well.

PR #629, opened 2019-03-25
tests/dockerfile: update base image

jkoelker:
Changes the base image to `mesosphere/marathon-lb:latest`. Since this
image requires a copy of `/marathon-lb`, reusing the existing image
instead of creating another one helps with caching. Also since the
latest `cassandra-driver` no longer works on `debian:jessie` and
`mesosphere/marathon-lb:latest` is already built upon `debian:buster`
using it as the base makes sense.

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -1,12 +1,21 @@
-FROM debian:jessie
+FROM mesosphere/marathon-lb:latest
 
-COPY  . /marathon-lb
-
-RUN apt-get update && apt-get install -y ruby apache2 vim curl ruby-dev build-essential \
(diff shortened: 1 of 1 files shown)


tests/Dockerfile
```diff
@@ -1,12 +1,21 @@
...
+        vim \
+        ruby-dev \
+        build-essential \
+        cassandra \
     && gem install --no-ri --no-rdoc cassandra-driver \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /marathon-lb
+
+# NOTE(jkoelker) Clear out the entrypont and cmd from the parent image
```
ichernetsky: Nit: entrypoint

PR #640, opened 2019-06-19
Use self.host property once in get_event_stream

sharego:
`self.host` cycles through available Marathon URLs, and it was accessed twice in `get_event_stream`, which made it access the same Marathon instance, if two URLs were supplied, which is especially profound if the instance being accessed over and over again is down.


jkoelker:
ok to test

ichernetsky:
@sharego, thanks a lot for your contribution and your patience. I reviewed the patch, and I will make sure that it gets merged in no time.

Issue #655, opened 2019-11-04
tasks are not passing health checks but tasks were receiving traffic


mslee89:
Hello.
I had an issue when I tried to deploy app on DC/OS.
The new tasks are not passing health check, but those tasks were receiving traffic before health check query.

My info
 - DC/OS 1.11.10
 - Marathon-LB 1.12.3

App info (DC/OS)
`{
  "labels": {
    "HAPROXY_GROUP": "external",
    "HAPROXY_0_FRONTEND_HEAD": "",
    "HAPROXY_0_FRONTEND_BACKEND_GLUE": "",
    "HAPROXY_0_VHOST": "test.com"
  },
  "id": "/test",
  "backoffFactor": 1.15,
  "backoffSeconds": 1,
  "container": {
    "portMappings": [
      {
        "containerPort": 5000,
        "servicePort": 10131,
        "name": "python"
      }
    ],
    "type": "DOCKER",
    "volumes": [],
    "docker": {
      "image": "python:2",
      "forcePullImage": false,
      "privileged": false,
      "parameters": []
    }
  },
  "cpus": 2,
  "disk": 0,
  "healthChecks": [
    {
      "gracePeriodSeconds": 300,
      "intervalSeconds": 60,
      "maxConsecutiveFailures": 3,
      "portIndex": 0,
      "timeoutSeconds": 60,
      "delaySeconds": 15,
      "protocol": "MESOS_HTTP",
      "path": "/health",
      "ipProtocol": "IPv4"
    }
  ],
  "instances": 3,
  "maxLaunchDelaySeconds": 3600,
  "mem": 2048,
  "gpus": 0,
  "networks": [
    {
      "name": "dcos",
      "mode": "container"
    }
  ],
  "requirePorts": false,
  "upgradeStrategy": {
    "maximumOverCapacity": 0.1,
    "minimumHealthCapacity": 1
  },
  "killSelection": "YOUNGEST_FIRST",
  "unreachableStrategy": {
    "inactiveAfterSeconds": 0,
    "expungeAfterSeconds": 0
  }
}`

Log 
```
 * Running on http://0.0.0.0:5000/ (Press CTRL+C to quit)
44.128.0.31 - - [04/Nov/2019 02:27:02] "GET / HTTP/1.1" 200 -
44.128.0.31 - - [04/Nov/2019 02:27:02] "GET / HTTP/1.1" 200 -

[...]

PR #656, opened 2019-11-08
Fix bug in checking healthChecks at first (#655)

chozo99:
Suppose you have an app that needs to warming up at first.
( e.g. http response code: 503 -> 200 )

When start a app that has healthChecks
task have healthChecks Results property but don't have any element in array at first

But it add backend server in haproxy

Eventually, 
Before healthChecks run once at first
request's packet will flow in a app that don't check healthChecks

#655

### DIFF
--- a/marathon_lb.py
+++ b/marathon_lb.py
@@ -1691,10 +1691,10 @@ def get_apps(marathon, apps=[]):
                 for task in new['tasks']:
                     if 'healthCheckResults' not in task:
                         continue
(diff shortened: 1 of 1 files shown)


jkoelker:
ok to test

jkoelker:
ok to test

jkoelker:
ok to test

jkoelker:
Thanks!

PR #659, opened 2019-11-28
Add prometheus-exporter to HAProxy

rfvbkm:
As of 2.0.0, HAProxy includes a Prometheus exporter module that can be built into a binary during build time.

To build with the official Prometheus exporter module, `make` should be run with the following `EXTRA_OBJS` flag:
`make TARGET=linux-glibc EXTRA_OBJS="contrib/prometheus-exporter/service-prometheus.o"`
Once built, prometheus-exporter can be enabled in haproxy.cfg:

```
listen stats
  bind 0.0.0.0:9090
  balance
  mode http
  stats enable
  http-request use-service prometheus-exporter if { path /metrics }
```
For more infromation, see this [official blog post](https://www.haproxy.com/blog/haproxy-exposes-a-prometheus-metrics-endpoint/)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,6 +81,7 @@ RUN set -x \
         USE_REGPARM=1 \
         USE_STATIC_PCRE=1 \
         USE_ZLIB=1 \
+	    EXTRA_OBJS="contrib/prometheus-exporter/service-prometheus.o" \
         all \
         install-bin \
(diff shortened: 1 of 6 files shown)


jkoelker:
ok to test

harryge00:
Will this be released in v1.14.3?

PR #667, opened 2020-04-02
docker: bump haproxy to 2.0.14, fix CVE-2020-11100

vixns:
https://www.mail-archive.com/haproxy@formilux.org/msg36877.html


jkoelker:
ok to test

PR #672, opened 2020-09-24
docker: bump haproxy to 2.0.17

jkoelker:


jkoelker:
Ugh, looks like CI aws credentials are broken, I built it locally and it builds and runs.

jkoelker:
nm, i forgot https://github.com/mesosphere/marathon-lb/pull/670 was there, going to bump the version in it, since it should also fix the ci ;)

